PR #6, opened 2023-06-30
improving prompt template

vipulgupta1011:


dmjoy:
Thanks @vipulgupta1011 , this LGTM.  You can just delete the previous return values that you have commented out (e.g. on line 222).  Once you make those changes I'll merge.

vipulgupta1011:
@dmjoy I have deleted the previous return values. I think we are good to go

PR #14, opened 2023-07-24
Viewed text retrieved by llama index and model output for different prompts

billray0259:


dmjoy:
Closing as we'd like to keep more experimental code in separate repositories.

PR #16, opened 2023-08-03
Refactor interfaces

dmjoy:
Major refactor to pull out the different interfaces (TA3 and local files for now) to be easier to use and pull into custom command line scripts.


brianhhu:
Two minor comments that don't necessarily have to be addressed for this PR:

1) Would it make sense to add the ITM-MVP server requirements as an "extras" requirement in poetry? That way everything can be installed uniformly using poetry, rather than setting up the base `align-system` using poetry, and requiring an additional step to install the TA3 server via the ITM-MVP Github instructions.

2) It might also make sense to include some "dummy" local files as a way to test that the overall system using the LocalFiles interface runs, as it's a slightly bigger lift to get the TA3 API installed just to test that the code works.

Otherwise, LGTM- thanks for refactoring this and making it easier for everyone to use!

brianhhu (commented):
See my other comment- thanks!

dmjoy:
Thanks for the feedback @brianhhu , I'm a little on the fence about include the TA3 Server install stuff as an extra set of dependencies (I think it's probably fine if it's neatly pip installable (which it is now), but if gets more involved than that, then I think it makes more sense to be a completely separate setup).

Regarding including some good sample data for testing the installation / setup, that's an excellent idea, if possible I'd like to just include the MVP scenarios from the TA1 folks, but maybe I should touch base with them before including them in a public repo?

PR #17, opened 2023-08-14
Added llama-2 option for llama_index algorithm

billray0259:


dmjoy:
LGTM @billray0259 , if you could update the `CHANGELOG.md` file to note the additional support for llama-2 that would be great (you probably need to make a new Version section at the top of the changelog, can call it `"Unreleased"` for now.

PR #28, opened 2023-10-19
Dev/language model lib

billray0259:


dmjoy:
I think it may make more sense to move this new `language_model_lib` code under `algorithms/chat` or something like that, and you could put the `align_system/language_model_lib/llama_2_kdma_predicting_adm.py` file directly under `algorithms`.  It may also make sense to rename `llm_chat_baseline.py` to `llama_2_chat_baseline.py` if it really is specific to llama 2.

PR #29, opened 2023-10-24
Dev/evaluation framework

billray0259:


billray0259:
Extract out the dataset loading logic form the evaluation code. If the evaluation code isn't holding any state, don't make it a class. Use the dataset loading logic of the interface code that already exists, just extend that to enable iterating functionality in standard format.

dmjoy:
Closing as these changes were covered in #32

PR #35, opened 2024-01-09
Paper experiment changes

billray0259:
Not all changes are merge-ready. Creating pull request draft to track changes with the goal of keeping main up-to-date.


align_system/algorithms/lib/chat/chat_language_model.py
```diff
@@ -53,26 +65,7 @@ def generate_responses(self,
             prefixes.append(prefix)
 
         # Tokenization step
-        try:
```
billray0259, line 56: Extracted this code out into `dialogs_to_token_list`

align_system/algorithms/llama_2_single_kdma_adm.py
```diff
@@ -13,29 +15,50 @@
...
+# default_system_messages_path=os.path.join(
+#     pathlib.Path(__file__).parent.absolute(), '..',
+#     'prompt_engineering/bbn_alignment_system_messages_v1')
+
+# NOTE temporary way to change which system messages are used
+# TODO make this configurable from the config
 default_system_messages_path=os.path.join(
     pathlib.Path(__file__).parent.absolute(), '..',
-    'prompt_engineering/bbn_alignment_system_messages_v1')
+    'prompt_engineering/single_kdma_adm_system_messges')
```
billray0259, line 57: If all the system messages are in the same directory then being able to configure them is less important... configuration could still be useful if we're exploring using different system messages for the same KDMA

PR #43, opened 2024-04-07
Fix minor bug in Llama-2 Single KDMA ADM When Answer is not integer

DavidMChan:
When the answer to a parsed treatment output is not an integer (can sometimes happen with the LLM), instead of failing on L959, try to parse the answer into an integer (and retry if impossible).


dmjoy:
Thanks David, LGTM.  Are you planning to make further changes on this particular MR?  If not I'll go ahead and merge

DavidMChan:
Nope! I just wanted to test that I had everything working! Thanks!

PR #44, opened 2024-04-16
Updated START tagging ADM

eveenhuis:
- Modified prompt
- Removed duplicated RED tag example, replaced with missing BLACK tag example
- Removed duplicated inference call in `identify_tag_color`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@
 This changelog follows the specifications detailed in: [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html), although we have not yet reached a `1.0.0` release.
 
+## Unreleased
+
+### Changed
+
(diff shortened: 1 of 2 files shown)


dmjoy:
@eveenhuis no issue here with the proposed changes.  Though please update the `CHANGELOG.md` file, if there isn't one already I typically create a new section above the latest release called `Unreleased` to track the changes prior to cutting an actual release, e.g.:
```
## Unreleased

### Added
- Added something

### Changed

### Fixed
- Fixed something

```

dmjoy:
Slightly longer term it is worth thinking about capturing this prompt and samples somewhere alongside the other templates/texts so that we can switch between them more easily if we ever need to (for a/b testing etc.)

eveenhuis:
@dmjoy Added the changelog entry

Completely agree about the templating, Bill already has that as a TODO in the code. It doesn't look like we really use issues in this repo, is there somewhere specific we should note that down?

dmjoy:
Thanks for the update, merging shortly.

We haven't historically used the issues here, but seems like a good time to start (especially since this is a very code specific task).

PR #46, opened 2024-04-17
Oracle ADM

eveenhuis:


align_system/algorithms/random_adm.py
```diff
@@ -1,6 +1,6 @@
 import random
 
-from swagger_client.models import ActionTypeEnum, InjuryLocationEnum, CharacterTagEnum
+from swagger_client.models import ActionTypeEnum, InjuryLocation, CharacterTagEnum
```
eveenhuis, line 3: @dmjoy, do you know what the guidance is on `InjuryLocationEnum` vs `InjuryLocation`? I was running into errors when `InjuryLocationEnum` was being used, but that looks like what we're using in the single KDMA ADM

dmjoy: I'm not certain, it looks like in the server code they're exactly the same apart from the name.. what error were you running into?

eveenhuis: `InjuryLocation` has less options than `InjuryLocationEnum` (and they seem to be validating against `InjuryLocation`), so I'd get an error when it chose an "invalid" location from `InjuryLocationEnum`

dmjoy: Ah, it looks like they've made some changes to the `development` branch of the client (and specifically the `InjuryLocationEnum` class) in the last few days.  Can you install the `main` branch of the client library instead?  I'll update the `pyproject.toml` file to this effect as well, but at the time that was THE branch to be using as we approached the metrics evaluation, but now `main` is the one we should target

align_system/algorithms/oracle_adm.py
```diff
@@ -0,0 +1,321 @@
...
+                return None
+
+            # Determine what treatment to apply, narrow down applicable characters
+            if 'treatment' not in action_to_take.parameters:
+                # Can only apply treatments if supplies are available
+                poss_treatments = [s.type for s in scenario_state.supplies if s.quantity > 0]
+                ta3_server_treatments = [
+                    'Burn Dressing', 'Decompression Needle', 'Hemostatic gauze', 'Nasopharyngeal airway',
+                    'Pressure bandage', 'Splint', 'Tourniquet', 'Pain Medications'
+                ]
```
dmjoy, line 163: Ideally these values would be pulled from an enum, I'm guess it's not in the case due to differences between the client enums and what's supported on the server?

align_system/algorithms/oracle_adm.py
```diff
@@ -0,0 +1,321 @@
...
+                    for c in poss_chars:
+                        for injury in c.injuries:
+                            potential_treatment = _get_treatment(poss_treatments, injury.name, injury.location)
+                            if potential_treatment is not None:
+                                tmp.append(c)
+                                would_be_used.add(potential_treatment)
+                                break
+                    poss_chars = tmp
+
+                if not would_be_used:
```
dmjoy, line 190: I've seen the convention a few times here of `if not some_enumerable_thing` to check if it's empty or not, but I think it's much more clear to say `if len(some_enumerable_thing) == 0`.  Otherwise at a glance I'm not sure if `would_be_used` is a bool or what (and have to check elsewhere in the code to confirm).

align_system/algorithms/oracle_adm.py
```diff
@@ -0,0 +1,321 @@
...
+                    log.warn("No treatments left, defaulting to random treatment")
+                    action_to_take.parameters['treatment'] = random.choice(poss_treatments)
+                else:
+                    action_to_take.parameters['treatment'] = random.choice(list(would_be_used))
+                log.info(f"Selected treatment: {action_to_take.parameters['treatment']}")
+
+            if 'location' not in action_to_take.parameters:
+                potential_locations = set()
+
+                # Limit possible characters to those with injuries that make sense for chosen treatment
```
dmjoy, line 200: I could be missing some potential condition, but would it simplify (or reduce) the code if when the action to take is an APPLY_TREATMENT action, pick a random character with an untreated injury first, and then pick one of the injuries and then choose the treatment and location to treat said injury (so more top-down rather than bottom-up?)

align_system/algorithms/oracle_adm.py
```diff
@@ -0,0 +1,321 @@
...
+            for action in available_actions
+        ]
+
+        # Convert inverse distances to probabilities
+        probs = [inv_dist/sum(inv_dists) for inv_dist in inv_dists]
+
+        if self.probabilistic:
+            action_to_take = np.random.choice(available_actions, p=probs)
+        else:  # Always choose max probability action (chooses first if multiple are tied)
+            action_to_take = available_actions[probs.index(max(probs))]
```
dmjoy, line 78: Tagging @brianhhu here to be sure, but I think Brian had intended the probabilistic piece here to be a knob to turn (something like 95% of the time it picks the best answer vs. 90% vs. 80% etc.) rather than use the distance from the target as probabilistic weights.  Brian could you confirm?

dmjoy (changes requested):
Overall this looks fine and don't have too much concern about merging as is assuming it's working.  But I do want to clear up the question about how the probabilistic piece is set up, and just consider my other bits of feedback.  Please also update the changelog.  Thanks!

align_system/algorithms/oracle_adm.py
eveenhuis: I guess that would work. My thought process was that we may not have the ability to treat the injury that ends up getting selected. However, I guess we could just try again if that was the case.

align_system/algorithms/oracle_adm.py
eveenhuis: Yeah, the probabilistic piece needs the most work still. Happy to hear any feedback/desires there

I thought Brian mentioned only smoothing probabilities without a matching KDMA, but in the current scenarios that doesn't really affect the final outcome (as actions without KDMAs are essentially no-ops right now). My "smoothing" technique is also not very technical right now, so that could likely be improved, just not sure how

align_system/algorithms/oracle_adm.py
dmjoy: Thanks Emily, I think it's a fair bit easier to follow now (at least to my eyes) and seems a fair bit shorter as well.

align_system/algorithms/oracle_adm.py
dmjoy: Again will wait for Brian to chime in here, but perhaps a simple way to implement what I was suggesting above would be to check `random.random() < threshold` (where `threshold` is `0.95` or `0.9` etc.) and if the check passes just pick the "right" answer, otherwise just pick a random "not right" answer.

brianhhu: The idea of the probabilistic action sampling is that it can potentially support a wide variety of decision strategies (all within the same framework):

1) random sampling would just be uniform distribution of probabilities over all actions
2) the "greedy" approach would just put all the probability mass on one action (i.e. "most aligned" given the target KDMA), or a small subset actions if there's a tie
3) in between the random/greedy strategies would be sampling probabilities of actions weighted by some distance from the target KDMA- this allows for some natural variability in decisions (approximating humans), where it's possible for a "suboptimal" decision to be chosen on a given probe (but on average, decisions would be aligned)

I'll go through the code in a bit more detail to see if I understand what's going on and can provide additional feedback then.

eveenhuis: I removed this `self.probabalistic` if condition in the latest, the portion of the code that's most relevant to this discussion is probably [here](https://github.com/ITM-Kitware/align-system/pull/46/files#diff-717c95a115392c6e4474060165d525403aa1a6ec36c071483540ab1b1d179ccdR75-R80)

1. I think this would just be the `RandomADM`, correct me if I'm wrong
2. I believe this is the `else` portion of the linked code
3. I believe this is the `if` portion of the linked code. I'm not sure if we need to add additional weight to non-optimal choices to get more probability of taking those paths.[ Upweighting missing KDMAs](https://github.com/ITM-Kitware/align-system/pull/46/files#diff-717c95a115392c6e4474060165d525403aa1a6ec36c071483540ab1b1d179ccdR39-R40) may or may not be enough for that

brianhhu: Sorry I lost track of this MR, and thanks for addressing my last comment. I don't see a drawback for merging, as @eveenhuis reported alignment scores in-line with the maximum possible reported by TA1. We should go back and also check the Parallax "maximally aligned" metric (which should be close to 100%), but that will require generating the output files using this ADM.

dmjoy (approved):

dmjoy:
Re Brian's latest feedback, no objections to merging from my end.  We can always revisit and tweak things later.  @eveenhuis any reason from your end to hold off on merging?  If not, I'll merge

eveenhuis:
Ready to merge from my end!

PR #63, opened 2024-07-10
Add Persona ADM

DavidMChan:
This PR will add the persona-based ADM. It currently runs, but is missing some key functionality. 

Known TODOs (before merge):
- [x] Update backstory sampling to sample KDMA-aligned backstories (right now uses random backstory sampling). This is easy to do when the backstories are labeled with KDMA values, but this still is a TODO pending some compute on our backend.
- ~~[ ] Update `populate_treatment_parameters`, `populate_tagging_parameters` and `generic_populate_character_id` to use backstory information.~~
- [x] Debug similarity parsing (can cause CUDA conflicts in some cases?)
- [x] Change logging to debug for many scenario/action prompts.
- [x] Debug if using chat model in a completion framework is causing issues (it doesn't seem like it, but maybe?)


DavidMChan:
Ran the first set of demographic probes on the backstories today (@dmjoy):

image

Interestingly, the first generation of backstories with demographics is showing that there is only a little bit of variation in the demographic probes for moral judgement and maximization. For moral judgement, many of the stories that select a 1 are inconsistent (and it's an LLM consistency sampling issue where instead of selecting 1 followed by 0 for 10, they selected one and then went on without the 0). For maximization, there is much more variation in the lower numbers, and the distribution is very uniform above 5 (but for some reason, nobody likes to select numbers less than 5 in either case).

dmjoy:
Hey David, thanks for the update.  Regarding moral judgement, I'm not terribly surprised as we had a really hard time convincing the LLMs (via system prompts) to choose low moral judgement actions (i.e. treating the perpetrator first all else being equal) and eventually had to sort of invert the choices (negatively counting votes for those choices) made with a high moral judgement to try and compensate.  Would also be interested to see what the demographics look like for the 3 new attributes (should be noted in the slack recently): "Ingroup Bias", "Quality of Life", "Value of Life".  Tagging @brianhhu here for his situational awareness.

DavidMChan:
I added two more commits today which do simple maximization/panel sampling, and with some new backstories, which have probe values for both maximization and moral judgement. For the other three new attributes, do we have a good definition prompt? For example, I use the following prompts for moral judgement and maximization:

> Moral judgment affects medical triage decisions by prioritizing morally deserving patients, like victims or helpers, over those responsible for the situation. High moral judgment considers intent and responsibility, while low moral judgment does not factor in moral deservingness. On a scale of 1 (would always prioritize responsible parties) to 10 (would always prioritize victims), how would you rate your moral judgment in medical triage decisions? Please give a numeric number between 1 and 10, followed by a period, then explain why.


> Maximization affects medical triage decision by prioritizing searching exhaustively to gather the most information possible to form their response and plan actions that achieve the maximum result. Satisficing, on the other hand, prioritizes finding a solution that is good enough to meet the minimum requirements as quickly as possible. On a scale of 1 (would always satisfice) to 10 (would always maximize), how would you rate your decision-making style in medical triage decisions? Please give a numeric number between 1 and 10, followed by a period, then explain why.

The last things that remain to be done are to (1) Use the backstories during the character selection process and (2) actually evaluate the system to get some metrics. Since we have about 5k backstories, and we're only using a panel size of 3, the decisions are somewhat under-powered, but without it getting pretty expensive, I'm not sure we can do better.

dmjoy:
@eveenhuis see David Chan's comment above about what kind of prompts / text is needed for the new attributes

dmjoy:
I was able to run this ADM using Llama3-8b seemingly without issue (though quite a few failure to parse errors, which should be a non-issue with the switch over to outlines), and I otherwise don't have any issues / concerns with the code.  Merging!

PR #66, opened 2024-07-18
Limiting reasoning length to avoid OOM error

jadie1:
Adding `maxLength` to reasoning string outputs in json schemas so that they don't get stuck in repetitive loops and result in out of memory errors. Also fixed a couple of typos in regression prompts.


dmjoy:
LGTM!  Please update the Changelog as well and I'll hit the merge button

PR #67, opened 2024-07-19
Add Outlines Sampler Parameter

dmjoy:
FYSA @jadie1 @eveenhuis , I'm going to merge this soon but this MR tweaks both outlines ADMs (regular and regression) to support different outlines samplers (see the example experiment included).


dmjoy:
Oh and @jadie1 it looks like the `itertools.batched` function I pointed you at before is only supported in Python 3.12 and newer (where we're intending to support 3.10 and newer) so I've replaced that call with a suggested alternative from the Python docs.

PR #71, opened 2024-07-24
Add dialog provider for personas project

DavidMChan:
This PR adds a "Persona Provider" which can be instantiated and queried to get dialogs relating to aligned personas. Right now, this *Should* behave almost identically to the personas ADM, but returns turns in the form needed by our updated ADMs such as the outlines ADM.

### DIFF
--- a/align_system/algorithms/lib/persona/__init__.py
+++ b/align_system/algorithms/lib/persona/__init__.py
@@ -1,0 +1,1 @@
+from .provider import PersonaProvider # noqa: F401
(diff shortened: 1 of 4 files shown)


dmjoy (approved):
Thanks for the quick turnaround on this David, I don't have any concerns / critical feedback here, looks like just what we discussed.  I'm happy to merge this unless you're planning additional changes on this PR

DavidMChan:
Sounds good - No additional changes planned, since it looks like what you need. Feel free to merge!

PR #72, opened 2024-07-24
Fix for batch size of 1

jadie1:
Fix for running outlines ADMs with a generator batch size of 1


dmjoy:
Thanks Jadie, is this fix also needed for the outltines regression ADM?

dmjoy:
Closing as these changes are included with #73

PR #75, opened 2024-07-25
Fix bug in personas provider caching

DavidMChan:
Previously, the persona provider didn't correctly cache backstories in a repeatable way. This corrects it so that:
- If "Cache" is True, then the backstories for the current alignment will be cached (or returned from an existing cache)
- If "Cache" is False, then the backstories for the current alignment target will be cleared, and refreshed.


eveenhuis:
Does this PR still need to be merged?

DavidMChan:
Ah, yes! It would be great if this could be merged!

PR #78, opened 2024-07-29
Updated backstories with new KDMAs

DavidMChan:
This PR updates the personas ADM with new KDMAS:
* Ingroup bias
* Value of life
* Quality of life


align_system/algorithms/lib/persona/provider.py
```diff
@@ -14,6 +14,16 @@
 KDMA_TO_PROBE_MAPPING = {
     'MoralDesert': 'moral_judgment',
     'maximization': 'maximization',
+    'Moral judgement': 'moral_judgment',
+    'ingroup_bias': 'ingroup_bias',
+    'IngroupBias': 'ingroup_bias',
+    'Ingroup bias': 'ingroup_bias',
+    'value_of_life': 'value_of_life',
+    'ValueOfLife': 'value_of_life',
+    'Value of life': 'value_of_life',
```
eveenhuis, line 23: VoL was recently [renamed "Perceived quantity of lives saved" (`PerceivedQuantityOfLivesSaved`)](https://nextcentury.atlassian.net/wiki/spaces/ITMC/pages/3214704665/SoarTech+Dry+Run+Evaluation#KDMAs), and the definition has also been modified slightly, unfortunately. Do you think we can we get away with changing the mapping here or do you think the backstories would be invalid?

DavidMChan: As long as the definitions haven't changed that much, then we can probably still use the existing backstories. Here is the prompt for this:

> Value of life affects medical triage decisions by prioritizing the perceived overall survival rates of multiple lives over the immediate needs of an individual. High value of life considers total lives saved, while low value of life focuses on the most immediate individual need. On a scale of 1 (would always prioritize the most immediate individual need) to 10 (would always prioritize perceived overall survival rates of lives), how would you rate your value of life in medical triage decisions? Please give a numeric number between 1 and 10, followed by a period, then explain why.

Does this probe still apply? Or do we need to update the probe?

eveenhuis: I think that probe still fits well, actually!

DavidMChan: Sounds good! Then the only thing that has to change is this mapping!

eveenhuis: Awesome! It'd be great if you could make that change and then this looks ready to merge in, if you're ready on your end

DavidMChan: Updated! I left the old mapping for backwards compatibility, but let me know if you want me to remove this!

PR #79, opened 2024-07-29
Comparative regression ADM

jadie1:
Adds another version of regression where KDMA scores are predicted for all responses/choices at once
Run by specifying `adm=outlines_regression_aligned_comparative`


jadie1:
I just remembered I don't have ICL working yet with comparative regression, so closing this PR for now

PR #80, opened 2024-07-29
Comparative regression, templates, and COT ICL

jadie1:
- Adds another version of regression where KDMA scores are predicted for all responses/choices at once. Run by specifying `adm=outlines_regression_aligned_comparative`
- Adds template option for regression ADMs which can be turned on by setting `kdma_score_examples: True`. Run by specifying `adm=outlines_regression_aligned/template` or `adm=outlines_regression_aligned_comparative/template`
- Adds in-context examples with chain of thought reasoning for Adept KDMA's to regression ADMs. Run by specifying `adm=outlines_regression_aligned/incontext` or `adm=outlines_regression_aligned_comparative/incontext`

### DIFF
--- a/align_system/algorithms/outlines_regression_adm.py
+++ b/align_system/algorithms/outlines_regression_adm.py
@@ -24,6 +24,7 @@ from align_system.prompt_engineering.outlines_prompts import (
     outcome_prediction_prompt,
     outcome_prediction_json_schema,
     kdma_score_prediction_system_prompt,
+    kdma_score_prediction_system_prompt_with_examples,
     kdma_score_prediction_prompt,
     kdma_score_prediction_json_schema,
     regression_alignment_system_prompt,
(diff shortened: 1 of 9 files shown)


align_system/algorithms/outlines_regression_adm.py
```diff
@@ -99,33 +159,61 @@ def sample_outcome_predictions(self,
...
+
+                    # TODO: Include outcome in ICL example?
+                    icl_prompt = kdma_score_prediction_prompt(
+                        icl_scenario_description, icl_choices, icl_choice, None, dset_kdma
+                    )
+
+                    icl_datasets[dset_kdma].append({
+                        "prompt": icl_prompt,
+                        "expected_value": label[dset_kdma],
+                        "reasoning": get_chain_of_thought_reasoning(target_kdma, action, state, icl_choice, label[dset_kdma])
```
eveenhuis, line 204: Nit: Every time we use the label value, we multiply it by 10, maybe we should just do it once here?

eveenhuis: Not worth holding up the PR, though. Merging!

PR #81, opened 2024-07-30
TA1/TA3 Server Updates

eveenhuis:
Added intend_action capability
Added Soartech alignment targets
Added some random ADM SoarTech experiments
Small changes for TA3 API compatability


jadie1 (approved):
This all looks good to me

PR #84, opened 2024-08-01
Followup prompt fixes

eveenhuis:
- Prevent overwriting action data when using followup prompts
- Add aid followup prompt for API changes


jadie1 (approved):
Looks ready to merge!

PR #85, opened 2024-08-01
Add Hybrid KDMA regression approach

barry-ravichandran:
This branch adds the KDMA regression approach based off of the https://github.com/ITM-Kitware/attribute-alignment/tree/classifier repo.


align_system/algorithms/outlines_hybrid_regression_adm.py
```diff
@@ -0,0 +1,215 @@
...
+
+                target_kdma_configs.append(
+                    {
+                        "name": target_kdma_name,
+                        "score": predictions[choice_index] * 10,
+                        "action_choice": final_choice
+                    }
+                )
+
+        log.info('Predicted KDMA value from Regression model = %f', predictions[choice_index])
```
eveenhuis, line 202: If we move this info to the previous log, we probably don't need to re-log the same info

align_system/algorithms/outlines_hybrid_regression_adm.py
```diff
@@ -0,0 +1,215 @@
...
+        action_to_take.justification = (
+            f"The direct kdma predictions from a pre-trained bert-based-uncased regression model"
+            f" for the chosen action - \"{final_choice}\" - is equal to {predictions[choice_index] * 10}."
+        )
+
+        # Set up simple diaolg to return for follow-ups
+        alignment_system_prompt = regression_error_alignment_system_prompt(target_kdma_configs)
+        prompt = action_selection_prompt(scenario_description, choices)
+        dialog = [{'role': 'system', 'content': alignment_system_prompt},
+                  {'role': 'user', 'content': prompt}]
```
eveenhuis, line 213: This should be the [same as in the LLM regression ADM](https://github.com/ITM-Kitware/align-system/blob/main/align_system/algorithms/outlines_regression_adm_comparative.py#L531-L534), additionally we should probably re-generate the scenario description to one that is not choice-specific:

`scenario_description = scenario_state_description_dre(scenario_state)`

align_system/prompt_engineering/outlines_prompts.py
```diff
@@ -607,3 +607,33 @@ def comparative_kdma_score_prediction_json_schema(choices):
...
+    {{ scenario_state.unstructured.rstrip() }}
+    {{ scenario_state.environment.decision_environment.unstructured }}
+    {% if character is not none %}
+      {{ character.name }} - {{ character.unstructured }}
+    {% endif %}
+    """
+
+
+@outlines.prompt
+def regression_error_alignment_system_prompt(target_kdmas):
```
eveenhuis, line 624: We won't be using this anymore, so reminder to please remove

align_system/prompt_engineering/outlines_prompts.py
```diff
@@ -607,3 +607,33 @@ def comparative_kdma_score_prediction_json_schema(choices):
         "required": list(choices)
     }
     return json.dumps(json_schema)
+
+
+@outlines.prompt
+def scenario_description_classic_regression(scenario_state, character):
```
eveenhuis, line 613: I named this poorly, can you rename this something that makes it clear it's for your ADM?

align_system/algorithms/outlines_hybrid_regression_adm.py
```diff
@@ -0,0 +1,215 @@
...
+    def top_level_choose_action(self,
+                                scenario_state,
+                                available_actions,
+                                alignment_target,
+                                num_samples=1,
+                                predict_outcomes=False,
+                                distribution_matching='average',
+                                generator_batch_size=5,
+                                kdma_descriptions_map='align_system/prompt_engineering/kdma_descriptions.yml',
+                                kdma_score_examples=False,
```
eveenhuis, line 136: Do you need any of these? If not please remove

Side note: I wonder if it makes sense to get multiple predictions from your model -- I think this is what Jadie was getting at when she asked how deterministic your approach is. If your model always predicts the same value, it wouldn't make sense to sample multiple time, but if there's variance we could sample multiple times and do something like average

align_system/algorithms/outlines_hybrid_regression_adm.py
```diff
@@ -0,0 +1,215 @@
...
+            model_name,
+            device=device,
+            model_kwargs=kwargs.get('model_kwargs', {}),
+            tokenizer_kwargs=kwargs.get('tokenizer_kwargs', {}))
+        # NOTE: In cases where we want multiple samples, we're passing
+        # in a list of prompts (this allows us to shuffle answers in
+        # each prompt), rather than setting the number of samples in
+        # the sampler itself (which defaults to 1); setting the number
+        # of samples in the sampler may result in unexpected behavior
+        self.sampler = sampler
```
eveenhuis, line 125: I think we should just let this be inherited if you don't need to do anything in particular for your approach here

eveenhuis:
Don't forget to add a changelog entry

jadie1 (commented):
@eveenhuis did a great job on the initial code review. I will look over it again after you've addressed her feedback. I just wanted to echo that if this ADM currently only works with single scalar targets, we should add a runtime error if the targets don't match that (or at least document it with comments)

align_system/algorithms/outlines_hybrid_regression_adm.py
```diff
@@ -0,0 +1,215 @@
...
+                choice_index = np.argmin(dist_to_align_score)
+                final_choice = choices[choice_index]
+
+                target_kdma_configs.append(
+                    {
+                        "name": target_kdma_name,
+                        "score": predictions[choice_index] * 10,
+                        "action_choice": final_choice
+                    }
+                )
```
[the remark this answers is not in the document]
eveenhuis: So you've made changes on this in the latest version of the code, but I think there's some confusion still. One singular action and/or target (in the future) may have multiple KDMA values.

So we might be aligning to `{"ingroup_bias": 0.7, "moral_judgement": 0.2}` (at the same time), therefore we'd predict both two (in this case) KDMA values for each action and then select the action that minimizes the distance from the target overall, i.e. the singular action that is closest to **both** ingroup_bias=0.7 and moral_judgement=0.2. Does that make sense?

**However**, Jadie just pushed [alignment utilities](https://github.com/ITM-Kitware/align-system/blob/main/align_system/utils/alignment_utils.py) that I think you can make use of to avoid re-implementing this logic and also support distribution-based targets. She uses these in the comparative regression ADM [here](https://github.com/ITM-Kitware/align-system/blob/main/align_system/algorithms/outlines_regression_adm_comparative.py#L371-L397)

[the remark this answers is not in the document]
jadie1: @barry-ravichandran Let me know if you'd like help integrating [alignment utilities](https://github.com/ITM-Kitware/align-system/blob/main/align_system/utils/alignment_utils.py) with your ADM

align_system/configs/adm/hybrid_regression.yaml
```diff
@@ -0,0 +1,11 @@
...
+  _target_: align_system.algorithms.outlines_hybrid_regression_adm.HybridRegressionADM
+
+  model_name: mistralai/Mistral-7B-Instruct-v0.2
+
+inference_kwargs:
+  models:
+    model_name: 'bert-base-uncased'
+    target_checkpoint:
+      Moral judgement: /data/users/barry.ravichandran/ITM/attribute-alignment/checkpoints_full_train/moral_judgement/moral_judgement_bert-base-uncased_3e-05_2_16_epoch_16_test_acc_0.004499295485639288.pt
+      Ingroup Bias: /data/users/barry.ravichandran/ITM/attribute-alignment/checkpoints_full_train/ingroup_bias/ingroup_bias_bert-base-uncased_3e-05_4_10_epoch_10_test_acc_0.012139826783467082.pt
```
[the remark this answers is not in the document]
dmjoy: Good question, I don't think they should live here, but I'm not certain where / how best to store them.  Need to think a bit more about it

align_system/algorithms/outlines_hybrid_regression_adm.py
```diff
@@ -0,0 +1,203 @@
...
+                                available_actions,
+                                alignment_target,
+                                **kwargs):
+
+        scenario = []
+        for action in available_actions:
+            for character in scenario_state.characters:
+                chosen_character = None
+                if character.id == action.character_id:
+                    chosen_character = character
```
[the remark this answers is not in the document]
eveenhuis:
```
for action in available_actions:
    chosen_character = None
    if action.character_id is not None:
        for character in scenario_state.characters:
            if character.id == action.character_id:
                chosen_character = character
                break
     scenario_description = ...
```

align_system/algorithms/outlines_hybrid_regression_adm.py
```diff
@@ -0,0 +1,237 @@
...
+            predicted_kdma_values = {}
+            log.info("ACTION CHOICES AND REGRESSION MODEL PREDICTIONS:")
+            for pred, choice in zip(predictions_per_kdma, choices):
+                # Only works for scalar alignment targets
+                dist = (target_kdma_value - pred)**2
+                predicted_kdma_values.update({
+                    choice: {
+                        target_kdma_name: pred
+                    }
+                })
```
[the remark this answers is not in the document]
eveenhuis: I just tested this and it looks like it would be overwritten, I **think** you want to do something like

```
if choice not in predicted_kdma_values:
    predicted_kdma_values[choice] = {}
predicted_kdma_values[choice][target_kdma_name] = pred
```

eveenhuis (approved):

jadie1 (approved):

align_system/configs/adm/hybrid_regression.yaml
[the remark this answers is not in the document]
barry-ravichandran: @dmjoy I've moved the model checkpoints to a shared path.

align_system/algorithms/adms.py
```diff
@@ -14,4 +15,5 @@
     'OracleADM': OracleADM,
     'OutlinesTransformersADM': OutlinesTransformersADM,
     'PersonaADM': PersonaADM,
+    'HybridRegressionADM': HybridRegressionADM
```
dmjoy, line 18: No change needed here, but just FYSA "registering" an ADM in this file is no longer necessary with the Hydra configuration framework we're using now.  I have an MR to remove this file and older configs, but haven't merged it yet

dmjoy:
LGTM!

PR #87, opened 2024-08-02
Add leave one out option for comparative regression ADM

eveenhuis:
Adds leave one out option for comparative regression ADM (conditioning only on scenario description) as a means to limit the effect of testing on training data


eveenhuis:
@jadie1 Can you PTAL before I kick off a ton of runs to see if this sort of matches what you were thinking? We can't use probe ID because we aren't provided that information

jadie1:
That looks good, I think using the `scenario_description` in place of a probe ID makes sense

PR #89, opened 2024-08-05
KDE Alignment Functionality

jadie1:
- Add option to specify `num_kde_samples` for multiple samples in sample-based KDE target alignment for oracle 
- Adds `max-likelihood` KDE target alignment option for oracle and comparative regression ADMs
- Switches regression-based ADMs to use baseline prompt as the starting point for follow-ups. The old regression-specific prompt did not work with KDE targets. Now the baseline and regression-based ADMs have the exact same follow-up procedure 
- Updates the ST training experiments to use both train scenarios

### DIFF
--- a/align_system/configs/adm/oracle.yaml
+++ b/align_system/configs/adm/oracle.yaml
@@ -5,5 +5,8 @@ instance:
   upweight_missing_kdmas: false
   filter_unlabeled_actions: false
   misaligned: false
+
+inference_kwargs:
   distribution_matching: sample
-  kde_norm: globalnorm
+  num_samples: 1
+  kde_norm: rawscores
--- a/align_system/utils/distribution_matching_utils.py
+++ b/align_system/utils/distribution_matching_utils.py
@@ -1,22 +1,46 @@
 from align_system.utils import kde_utils
+import numpy as np
 
-
-def match_to_target_kde_sample(alignment_target, available_actions, kde_norm):
+def match_to_target_kde_sample(alignment_target, available_actions, kde_norm, num_kde_samples=1):
 	'''
-	Samples a random value from the target KDE and selects the action with KDMA value closest to the sample
+	Returns the choice with min average distance to random samples from the target KDE
 	'''
 	target_kdma = alignment_target.kdma_values[0] # TODO extend this to multi-KDMAs
 	target_kde = kde_utils.load_kde(target_kdma, kde_norm)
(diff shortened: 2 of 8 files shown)


align_system/utils/distribution_matching_utils.py
```diff
@@ -1,22 +1,46 @@
...
 
+	return selected_action
+
+
+def max_likelihood_matching(alignment_target, available_actions, kde_norm):
+	'''
+	MLE - Returns the choice with the max likelihood under the target KDE
+	'''
+	selected_action = None
+	target_kdma = alignment_target.kdma_values[0] # TODO extend this to multi-KDMAs
```
eveenhuis, line 33: I know we keep leaving these as TODOs, but I'm wondering where we're at the point where we're starting to get so many that we should actually check to confirm that we haven't been provided a multi-kdma target so that we aren't giving ourselves a footgun

align_system/configs/adm/oracle.yaml
```diff
@@ -1,7 +1,9 @@
 instance:
   _target_: align_system.algorithms.oracle_adm.OracleADM
 
+inference_kwargs:
   probabilistic: false
   upweight_missing_kdmas: false
```
eveenhuis, line 6: `probabilistic` and `upweight_missing_kdmas` are currently set up as instance `kwargs` (arguments of the constructor not `choose_action`). It might make sense to change them to inference kwargs, but I'm not sure this config is quite right as is (though works for this configuration because the config values match the defaults).

jadie1:
@eveenhuis I did not see your misaligned oracle PR. I'll review that and we can get it merged, then I'll update this PR to integrate it

align_system/utils/distribution_matching_utils.py
jadie1: Good point, I added a runtime error

align_system/configs/adm/oracle.yaml
jadie1: Thanks, I've fixed it

align_system/utils/distribution_matching_utils.py
eveenhuis: Do we need to do the same thing for any other ADMs?

jadie1: Just added for the comparative regression ADM

eveenhuis (approved):

PR #90, opened 2024-08-05
Fix for JS distance NaN value when KDEs are close

jadie1:
Adds fix for JS distance NaN value when KDEs are close, following https://github.com/ITM-Soartech/ta1-server-mvp/commit/cfb0c408b7b5fab687d26d13a99ff374b7f79139

### DIFF
--- a/align_system/utils/kde_utils.py
+++ b/align_system/utils/kde_utils.py
@@ -173,9 +173,16 @@ def js_distance(kde1, kde2, samples: int):
     x = np.linspace(0, 1, samples)
     pdf_kde1 = _kde_to_pdf(kde1, x)
     pdf_kde2 = _kde_to_pdf(kde2, x)
-    
-    # Compute the Jensen-Shannon Distance using samples
-    js = jensenshannon(pdf_kde1, pdf_kde2)
+
+    if np.allclose(pdf_kde1, pdf_kde2):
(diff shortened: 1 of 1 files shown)


align_system/utils/kde_utils.py
```diff
@@ -173,9 +173,16 @@ def js_distance(kde1, kde2, samples: int):
...
     pdf_kde2 = _kde_to_pdf(kde2, x)
-    
-    # Compute the Jensen-Shannon Distance using samples
-    js = jensenshannon(pdf_kde1, pdf_kde2)
+
+    if np.allclose(pdf_kde1, pdf_kde2):
+        # If two kdes are functionally identical but off by a 10 to the minus 6 or so floating point amount
+        # jensenshannon can hit floating point roundoff problems and return a nan instead of a zero.
+        # To avoid introducing nans by hitting this case, we'll set very close to zero cases to zero.
+        print("KDE1 and KDE2 are so close to identical Jensenshannon can produce a nan. Assigning them to be identical.")
```
eveenhuis, line 181: Can we change this to a log instead of a print, so that it'll show up in logs properly?

jadie1: I just removed it. I don't think we even need to log it, I had just copied that over. If the PDFs are identical the JS should be 0

PR #92, opened 2024-08-08
Add choose_action timing configuration option

eveenhuis:
Adds capability and configuration option to time the `choose_action` method and create a report summarizing those times.

### DIFF
--- a/align_system/configs/action_based.yaml
+++ b/align_system/configs/action_based.yaml
@@ -12,6 +12,6 @@ save_log: true
 save_input_output: true
 save_scoring_output: true
 save_alignment_targets: false
-save_timing: false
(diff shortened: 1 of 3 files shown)


align_system/configs/action_based.yaml
```diff
@@ -12,5 +12,6 @@ save_log: true
 save_input_output: true
 save_scoring_output: true
 save_alignment_targets: false
+save_timing: false
```
jadie1, line 15: Can we make this default true?

eveenhuis: Done!

jadie1:
Tried it out, looks good! Ready to merge

PR #94, opened 2024-08-13
Character info promopt

jadie1:
This PR adds a scenario prompt that includes all structured character information that is not identical between characters. 
I accidentally merged this functionality to main already (https://github.com/ITM-Kitware/align-system/commit/8aa795afdd53372dadb568732363e4730e41d3cf), but in this PR I adds a bug fix for missing keys and updates the change log.


dmjoy:
LGTM, and I looked over the commit already pushed to main, no complaints here.  Merging

PR #95, opened 2024-08-13
Don't allow treatments when supplies aren't available

eveenhuis:
Update ADEPT train experiments to include new scenarios. The new scenarios identified the need to restrict pre-specified treatments when supplies are insufficient, so a fix for this is included.


dmjoy (approved):
LGTM; no objection to merging from my side unless you want to build a few more changes on top

PR #96, opened 2024-08-13
SoarTech ICL

jadie1:
This PR adds the template and incontext learning chain of thought reasoning for SoarTech KDMAS (this is already covered in the changelog).

Additionally it fixes the structured character info included in the prompt to only include characters that have an action associated with them (the scenario states for SoarTech were including all characters).


jadie1:
Looks like SoarTech fixed this issue with extra characters https://inthemoment-group.slack.com/archives/C05AQMCA5PB/p1723499930664379
I don't think the extra logic in the code to filter out extra characters hurts anything though

align_system/utils/outlines_prompts_utils.py
```diff
@@ -85,14 +85,20 @@ def filter_dict(d, common):
     return [filter_dict(d, common_values) for d in dicts]
 
 
-def get_relevant_structured_character_info(characters):
+def get_relevant_structured_character_info(state, actions):
```
eveenhuis, line 88: I don't think we can restrict the characters if an open-ended action with an unspecified character ID is available, e.g TAG_CHARACTER. Does this cover that case?

align_system/algorithms/outlines_regression_adm.py
```diff
@@ -40,60 +40,96 @@ def get_chain_of_thought_reasoning(target_kdma, action, state, choice, label):
...
+                cot_reasoning += f"Because {action_character.name} would likely not survive treatment"
+                # Sub-optimal treatment
+                if expected_value == 3:
+                    cot_reasoning += " but the supplies needed to treat the other casualty would still be available"
+                cot_reasoning += ", choosing "
 
     else:
         cot_reasoning = f"Choosing "
+        
     cot_reasoning += f"{choice} would score a {expected_value} for the the principle of {target_kdma['name']}."
```
eveenhuis, line 128: I think this is an unintended unindent. Instead of appending, you may just want to continue the string on the next line with `\`

align_system/algorithms/outlines_regression_adm.py
```diff
@@ -40,60 +40,96 @@ def get_chain_of_thought_reasoning(target_kdma, action, state, choice, label):
...
+                cot_reasoning += ", choosing "
+
+    # Percieved quanity of lives saved
+    # Example:
+        # Because Casualty G would likely survive treatment, choosing Treat Casualty G with hemostatic gauze would score a 9 for the the principle of perceived quantity of lives saved.
+        # Because Casualty G would likely survive treatment but it is not the best treatment approach, choosing Treat Casualty G with pressure bandage would score a 7 for the the principle of perceived quantity of lives saved.
+        # Because Casualty H would likely not survive treatment but the supplies needed to treat the other casualty would still be available, choosing Treat Casualty H with pressure bandage would score a 3 for the the principle of perceived quantity of lives saved.        
+        # Because Casualty H would likely not survive treatment, choosing Treat Casualty H with hemostatic gauze would score a 1 for the the principle of perceived quantity of lives saved.
+    elif target_kdma['kdma'] == 'PerceivedQuantityOfLivesSaved':
+        if action_character is not None:
```
eveenhuis, line 109: What if action_character **is** None (applies to other KDMAs too)? I think `cot_reasoning` currently ends up being undefined in that case (if the indentation gets fixed)

align_system/algorithms/outlines_regression_adm.py
jadie1: The unindent is intentional but may be unclear, so I fixed here: https://github.com/ITM-Kitware/align-system/pull/96/commits/e0d0bbc2589d77949a02944ba7c3292db9297a76

If the choice is associated with a character, then the cot will be like: "Because {character} is... choosing, {choice} would score a {expected_value} for the the principle of {target_kdma['name']}"
If there is no character, then it will just be: "Choosing {choice} would score a {expected_value} for the the principle of {target_kdma['name']}"
Either way the reasoning will always end with "{choice} would score a ..."

align_system/utils/outlines_prompts_utils.py
jadie1: Good point, I don't think it'd cover that case. Should I take this change out and should we just rely on TA1 to include the correct set of characters?

align_system/algorithms/outlines_regression_adm.py
jadie1: Fixed here https://github.com/ITM-Kitware/align-system/commit/e0d0bbc2589d77949a02944ba7c3292db9297a76

align_system/utils/outlines_prompts_utils.py
eveenhuis: Yeah, I think we should take it out, so we're not making too many assumptions

jadie1:
@eveenhuis I removed the commits for filtering out characters

eveenhuis (approved):

dmjoy:
LGTM, longer term I would like to think of ways to move ICL and COT reasoning stuff to it's own module

PR #97, opened 2024-08-15
Update Hybrid Kaleido ADM for DRE

dmjoy:
Updated Hybrid Kaleido ADM to use the Outlines ADM for action parameter completion, and update to make use of KDE based alignment utils for SoarTech scenarios. (WIP)

### DIFF
--- a/align_system/algorithms/kaleido_adm.py
+++ b/align_system/algorithms/kaleido_adm.py
@@ -149,7 +149,13 @@ class KaleidoADM(AlignedDecisionMaker, ActionBasedADM):
 
         return results
 
-    def force_choice(self, kaleido_results, choices, distance_fn=DefaultDistanceFunction):
+    def force_choice(self, kaleido_results, choices, target_kdma_values, distance_fn=DefaultDistanceFunction):
+        # Make sure we don't modify the original when we add in the targets
+        kaleido_results = kaleido_results.copy(deep=True)
+
+        kaleido_results['target'] = kaleido_results['KDMA'].apply(
+            lambda k: target_kdma_values[k])
+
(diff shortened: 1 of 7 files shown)


dmjoy:
This PR is still a WIP, but not expecting any more changes to the outlines ADM.  Wanted to give you all (@eveenhuis @jadie1 @barry-ravichandran ) a chance to see specifically the outlines ADM changes to make sure your ADMs will not be affected.  My intent here was just to move the action parameter population pieces into their own methods so that I could re-use them in the Hybrid Kaleido ADM

align_system/algorithms/kaleido_adm.py
```diff
@@ -276,25 +309,75 @@ def choose_action(self, scenario_state, available_actions, alignment_target, **k
             decision_environment=scenario_state.environment.decision_environment.unstructured.strip(),
             characters_str=characters_str)
 
-        # Scaling KDMA values by 10 (range should be 0-10)
-        if not isinstance(alignment_target, dict):
-            alignment_target = alignment_target.to_dict()
+        choices_unstructured = [a.unstructured for a in available_actions]
```
eveenhuis, line 312: Do we need to ensure these choices are unique like we do for Outlines?

eveenhuis (commented):
LGTM other than my one minor question

align_system/algorithms/kaleido_adm.py
dmjoy: Ah good catch, yes I think that makes sense.  Looks like just using the outlines_adm `format_choices` function should cover it yes?

eveenhuis: I think so

PR #98, opened 2024-08-15
Comparative regression justification update

jadie1:
I've updated the comparative regression justification to include the outcome prediction (if on) and the score reasoning from the best sample. To do this I had to update the alignment functions class to optionally return the index of the best sample.

### DIFF
--- a/align_system/utils/alignment_utils.py
+++ b/align_system/utils/alignment_utils.py
@@ -86,16 +86,19 @@ class AvgDistScalarAlignment(AlignmentFunction):
         selected_choice, probs = self._select_min_dist_choice(choices, distances, misaligned)
 
         if return_best_sample_index:
-            best_sample_index = 0
-            min_dist = float('inf')
-            for sample_idx in range(len(samples)):
-                dist = 0
-                for target_kdma in target_kdmas:
-                    sample = kdma_values[selected_choice][target_kdma['kdma']][sample_idx]
-                    dist += _euclidean_distance(target_kdma['value'], sample)
-                if dist < min_dist:
-                    min_dist = dist
-                    best_sample_index = sample_idx
+            sample_distances = []
+            sample_indices = range(len(samples))
+            if len(sample_indices) == 1:
+                best_sample_index = 0
+            else:
+                # For the selected choice, find the sample closest to the target
+                for sample_idx in sample_indices:
+                    sample_dist = 0
+                    for target_kdma in target_kdmas:
+                        sample = kdma_values[selected_choice][target_kdma['kdma']][sample_idx]
+                        sample_dist += _euclidean_distance(target_kdma['value'], sample)
+                    sample_distances.append(sample_dist)
+                best_sample_index, _ = self._select_min_dist_choice(sample_indices, sample_distances, misaligned)
             return selected_choice, probs, best_sample_index
         else:
             return selected_choice, probs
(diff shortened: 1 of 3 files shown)


align_system/utils/alignment_utils.py
```diff
@@ -80,11 +85,24 @@ def __call__(self, kdma_values, target_kdmas, misaligned=False):
...
+            min_dist = float('inf')
+            for sample_idx in range(len(samples)):
+                dist = 0
+                for target_kdma in target_kdmas:
+                    sample = kdma_values[selected_choice][target_kdma['kdma']][sample_idx]
+                    dist += _euclidean_distance(target_kdma['value'], sample)
+                if dist < min_dist:
+                    min_dist = dist
+                    best_sample_index = sample_idx
+            return selected_choice, probs, best_sample_index
```
eveenhuis, line 99: This may be me not being capable of math/logic this late in the day, but is this (and alternatives) not doing essentially the same thing as `_select_min_dist_choice` to find the index? Should that function just return an index? I'm also not sure if this logic works for misaligned

I'm also not sure I'm sold on the variable number of returns, but I feel like @dmjoy would have thoughts on that

jadie1: It is super similar but different. At first we are trying to select the **choice** with the min distance by aggregating over the samples. Once the choice is selected, we have to find which **sample** had the min distance for that choice because that is the "best" sample and we want to use it's reasoning as justification. Does that make sense? I don't love all the almost identical looking but different logic in the code, but I'm not sure how else to do it.

eveenhuis: Ah yes, gotcha

jadie1: I don't have strong feelings about the variable number of returns. If that is something we typically avoid we can make it always return the best_sample_index. I just made it an option because it doesn't make much sense in the context of the oracle which doesn't have samples so will always return 0

jadie1: I don't think this will work for misaligned... Maybe there is a way to reuse the _select_min_dist logic. I'll work on it

eveenhuis: If we know what the distance of the selected choice was, we just need to choose the one closest to *that* if that makes sense

jadie1: If I'm following, you're suggesting selecting the sample with distance closest to the distance of the choice (aggregate over samples), which would give us sort of an average sample. I think we do want the best sample here - or closest to the target

jadie1:
@eveenhuis I switched the best sample selection to also use `_select_min_dist_choice()` so that the misaligned option works

eveenhuis:
Thanks! I will approve and leave the return value decision up to @dmjoy

eveenhuis (approved):

align_system/utils/alignment_utils.py
jadie1: I think that would work, I'll try it

jadie1: I've moved it, I think it is cleaner as a separate function. Let me know what you think https://github.com/ITM-Kitware/align-system/pull/98/commits/9f5b2acc462e64cbfc3e4e90b8a6c77541d837b3

align_system/utils/alignment_utils.py
```diff
@@ -79,8 +85,23 @@ def __call__(self, kdma_values, target_kdmas, misaligned=False):
             distances.append(distance)
 
         selected_choice, probs = self._select_min_dist_choice(choices, distances, misaligned)
-
         return selected_choice, probs
+    
+    def get_best_sample_index(self, kdma_values, target_kdmas, selected_choice, misaligned=False, kde_norm=None):
```
dmjoy, line 90: is the `kde_norm` argument being used here?

align_system/utils/alignment_utils.py
```diff
@@ -91,16 +112,21 @@ def __call__(self, kdma_values, target_kdmas, misaligned=False, kde_norm='global
...
+            self.sampled_target_kdmas.append(sampled_target_kdma)
 
         # Use avergae distance to sampled scalar targets
         AlignmentFunc = AvgDistScalarAlignment()
-        return AlignmentFunc(kdma_values, sampled_target_kdmas, misaligned)
+        return AlignmentFunc(kdma_values, self.sampled_target_kdmas, misaligned=misaligned)
+    
+    def get_best_sample_index(self, kdma_values, target_kdmas, selected_choice, misaligned=False, kde_norm=None):
+        # Use avergae distance to sampled scalar targets
+        AlignmentFunc = AvgDistScalarAlignment()
```
dmjoy, line 128: I think lowercase underscore separated is more standard for variables (I can maybe understand using it for `AlignmentFunc = AvgDistScalarAlignment` but since you're instantiating `AvgDistScalarAlignment` with `()`, it's an object now not a class)

dmjoy (approved):
I think this looks a bit better overall, but I think there's still probably some refactoring that could be done to make this code easier to follow and a bit more modular.  That said, I'm not trying to be too picky ahead of the deadline here so I'm content to merge this as is (unless some of the suggestions are quick fixes) and we can work on the refactors / improvements beyond the delivery

align_system/utils/alignment_utils.py
jadie1: Fixed here https://github.com/ITM-Kitware/align-system/pull/98/commits/e88a7524b644ebdadb0448ef5aa4fadf7d67a8c3

align_system/utils/alignment_utils.py
jadie1: It is not, fixed here https://github.com/ITM-Kitware/align-system/pull/98/commits/f063db1e7029ec986df50496491e0b6902854656

PR #99, opened 2024-08-17
Eval scenario fixes

eveenhuis:
Fixes to get a complete baseline run through current eval scenarios


eveenhuis:
@dmjoy ready for your review! I think this is already covered by "Small updates to Outlines ADM to be compatible with API updates" in the changelog, but let me know if you want me to add something else.

PR #106, opened 2024-08-22
Add probabilistic alignment option to alignment utils

eveenhuis:
Exposes probabilistic alignment option in oracle, comparative regression, and hybrid regression ADMs.

Small clean up for no longer used oracle configurations


eveenhuis:
FYSA: I would like Jadie's input on this PR so I'm not requesting reviews until she's back

align_system/utils/alignment_utils.py
```diff
@@ -169,21 +173,21 @@ def __call__(self, kdma_values, target_kdmas, misaligned=False, kde_norm='global
...
+            total_likelihood = 0.
             for target_kdma in target_kdmas:
                 if isinstance(target_kdma, KDMAValue):
                     target_kdma = target_kdma.to_dict()
 
                 target_kde = kde_utils.load_kde(target_kdma, kde_norm)
                 predicted_samples = kdma_values[choice][target_kdma.kdma]
                 log_likelihoods = target_kde.score_samples(np.array(predicted_samples).reshape(-1, 1))
-                total_likelihood = np.sum(np.exp(log_likelihoods))
+                total_likelihood += np.sum(np.exp(log_likelihoods))
```
eveenhuis, line 184: @jadie1 I would especially like you to review that this particular change isn't wrong. I *think* I caught a bug, but may be misunderstanding. Wouldn't have affected our current results since it's only in play with multi KDMAs

align_system/algorithms/oracle_adm.py
```diff
@@ -17,30 +15,15 @@
...
-        self.upweight_missing_kdmas = upweight_missing_kdmas
-        self.filter_unlabeled_actions = filter_unlabeled_actions
+    def __init__(self, misaligned: bool=False, **kwargs):
         self.misaligned = misaligned
 
     def choose_action(self, scenario_state, available_actions, alignment_target,
-                      distribution_matching: str='sample',
-                      kde_norm: str='rawscores', **kwargs):
+                      distribution_matching='sample', kde_norm='rawscores',
+                      probabilistic=False **kwargs):
```
jadie1, line 23: Missing a comma after probabilistic

align_system/utils/alignment_utils.py
```diff
@@ -27,7 +27,7 @@
 '''
 class AlignmentFunction(ABC):
     @abstractmethod
-    def __call__(self, kdma_values, target_kdmas, misaligned=False, kde_norm=None):
+    def __call__(self, kdma_values, target_kdmas, misaligned=False, kde_norm=None, probabilistic=False):
```
jadie1, line 30: Can we add `probabilistic` to the inputs description in the comment describing AlignmentFunction

align_system/utils/alignment_utils.py
```diff
@@ -93,10 +97,10 @@ def __call__(self, kdma_values, target_kdmas, misaligned=False):
                 distance += _euclidean_distance(target_kdma['value'], average_score)
             distances.append(distance)
 
-        selected_choice, probs = self._select_min_dist_choice(choices, distances, misaligned)
+        selected_choice, probs = self._select_min_dist_choice(choices, distances, misaligned, probabilistic=probabilistic)
         return selected_choice, probs
 
-    def get_best_sample_index(self, kdma_values, target_kdmas, selected_choice, misaligned=False):
+    def get_best_sample_index(self, kdma_values, target_kdmas, selected_choice, misaligned=False, **kwargs):
```
jadie1, line 103: Why was `**kwargs` added here?

align_system/utils/alignment_utils.py
jadie1: @eveenhuis I think you did catch a bug. This fix looks correct, we want to total likelihood of each choice to be the sum across all target KDMAs

align_system/algorithms/oracle_adm.py
eveenhuis: Fixed in [7f146b3](https://github.com/ITM-Kitware/align-system/pull/106/commits/7f146b39609076b556f247f42c774ce6863d2269)

align_system/utils/alignment_utils.py
eveenhuis: Fixed in [7f146b3](https://github.com/ITM-Kitware/align-system/pull/106/commits/7f146b39609076b556f247f42c774ce6863d2269)

align_system/utils/alignment_utils.py
eveenhuis: Whoops, removed in [7f146b3](https://github.com/ITM-Kitware/align-system/pull/106/commits/7f146b39609076b556f247f42c774ce6863d2269)

jadie1 (approved):

align_system/utils/alignment_utils.py
```diff
@@ -53,11 +55,15 @@ def _select_min_dist_choice(self, choices, dists, misaligned=False):
...
-        selected_choice = choices[random.choice(max_actions)]
+        if probabilistic:
+            selected_choice = np.random.choice(choices, p=probs)
+        else:
+            max_prob = max(probs)
+            max_actions = [idx for idx, p in enumerate(probs) if p == max_prob]
+            # Randomly chose one of the max probability actions
+            # TODO could add some tie breaking logic here
+            selected_choice = choices[random.choice(max_actions)]
+
```
dmjoy, line 66: Not asking for any changes here, but just want to draw a little bit of attention to this being a case where having two separate functions for determining the choice (min distance or probabilistic) that gets passed in somewhere else would make it so that we don't have to keep extending the argument list here with additional flags (and what happens when we want to add another kind of flavor, and then what's the interaction between them etc.) in addition to making each individual piece easier to reason about (and write unit tests for if we decide that's necessary).

dmjoy (approved):

align_system/algorithms/oracle_adm.py
```diff
@@ -17,30 +15,15 @@
...
 
 class OracleADM(ActionBasedADM):
-    def __init__(
-        self, probabilistic: bool=False, upweight_missing_kdmas: bool=False,
-        filter_unlabeled_actions: bool=False, misaligned: bool=False, **kwargs
-    ):
-        self.probabilistic = probabilistic
-        self.upweight_missing_kdmas = upweight_missing_kdmas
-        self.filter_unlabeled_actions = filter_unlabeled_actions
+    def __init__(self, misaligned: bool=False, **kwargs):
```
dmjoy, line 18: Missed this in my "official" review, but several of these init arguments (probabilistic etc.) are ignored now during initialization (and are now expected to be passed in for the inference_kwargs is that right?), do any of the config files need to be updated to reflect this change?  Also if we're not actually using the `**kwargs` in the init method I think we should remove it since folks will at least get an error if they pass in extraneous arguments

align_system/utils/alignment_utils.py
eveenhuis: I agree. It also felt silly that I had to add the option to every alignment function in two places (essentially)

align_system/algorithms/oracle_adm.py
eveenhuis: I just updated all the configs of the adms I touched. 

For the oracle in particular, we lost the `upweight_missing_kdmas`  (no longer an option) and `filter_unlabeled_actions` (always on) functionality when we switched to the separate alignment utils. So this was fully removing those since they aren't true options anymore

When I was rebasing this I realized I had switched probabilistic to an inference kwarg, when you were just talking about switching things over to instance args instead, so I can make that change if you prefer

dmjoy: Gotcha.  Yeah slight preference for moving it (`probabilistic`) back to the init arguments.  Thanks

eveenhuis: Done!

PR #114, opened 2024-08-28
Clamp KDE target samples

jadie1:
This is a tiny PR that just ensures the values sampled from target KDEs are between 0 and 1


dmjoy (approved):
LGTM, merging

PR #115, opened 2024-08-30
ICL Reformat

jadie1:
This PR has the following reformatting:
- Moves the `format_choices()` function from the `OutlinesTransformersADM` class in `outlines_adm.py` to a new utils file: `adm_utils.py`. This function is used across ADMs and in ICL so it's better as a standalone function. 
- Moves the ICL functionality to `incontext_utils.py`

I created the abstract class `IncontextExampleGenerator()` with an abstract method `set_icl_datasets()` that is called in initalization. The `set_icl_datasets()` function creates a dataset of possible ICL examples for each target KDMA, where each example is a dictionary with keys `prompt` and `response`. With this setup, we can create different implementations of `IncontextExampleGenerator()` (as I did with `ComparativeRegressionIncontextExampleGenerator()`) that have specific prompts and responses. 

Then `IncontextExampleGenerator()` has a generic function `select_icl_examples(target_kdma, prompt_to_match)` that selects a subset of the examples for the target KDMA given the current probe (`prompt_to_match`). This is where the `random` and `bert_similarity` down-selection strategies are implemented, and we can add more. 

So when an ADM uses ICL, we first initialize a specific instance of `IncontextExampleGenerator()`. Then for each prompt, we call  `generator.select_icl_examples()` to get the relevant examples of `prompt` and `response` pairs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,23 +3,31 @@
 This changelog follows the specifications detailed in: [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html), although we have not yet reached a `1.0.0` release.
 
-## 0.5.2
+## Unreleased
 
 ### Changed
 
 * Moved incontext learning functionality into `incontext_utils.py` and updated the base outlines and comparative regession ADMS to use this module.
-* Moved the `format_choices()` function from the `OutlinesTransformersADM` class in `outlines_adm.py` to a new utils file: `adm_utils.py` so it can be used across ADMs. 
+* Moved the `format_choices()` function from the `OutlinesTransformersADM` class in `outlines_adm.py` to a new utils file: `adm_utils.py` so it can be used across ADMs.
 
 ### Added
 
 * Added option to normalize KDMA values in incontext examples
+  
+### Fixed
+
+* Fixed KDE target samples to be between 0 and 1
+  
+## 0.5.2
+
+### Added
+
 * Split out our experiment configuration for our aligned DRE ADM to specific configs for SoarTech and Adept
 * Added logging for sampled KDMA target value, and estimated KDMA values in alignment_utils
 
 ### Fixed
 
 * Fixed issue in Oracle ADM which caused an key error exception when logging probabilities
-* Fixed KDE target samples to be between 0 and 1
 
 ## 0.5.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,14 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## 0.5.2
 
+### Changed
+
+* Moved incontext learning functionality into `incontext_utils.py` and updated the base outlines and comparative regession ADMS to use this module.
+* Moved the `format_choices()` function from the `OutlinesTransformersADM` class in `outlines_adm.py` to a new utils file: `adm_utils.py` so it can be used across ADMs. 
+
 ### Added
 
+* Added option to normalize KDMA values in incontext examples
 * Split out our experiment configuration for our aligned DRE ADM to specific configs for SoarTech and Adept
 * Added logging for sampled KDMA target value, and estimated KDMA values in alignment_utils
 
--- a/align_system/algorithms/outlines_adm.py
+++ b/align_system/algorithms/outlines_adm.py
@@ -3,6 +3,7 @@ import random
 import itertools
 import numpy as np
 import torch
+import yaml
 
 import outlines
 from outlines.samplers import MultinomialSampler
@@ -17,6 +18,7 @@ from swagger_client.models import (
 
 from align_system.utils import logging
 from align_system.utils import adm_utils
+from align_system.utils import incontext_utils
 from align_system.utils import get_swagger_class_enum_values
 from align_system.utils.voting import (
     calculate_votes,
@@ -106,7 +108,7 @@ class OutlinesTransformersADM(ActionBasedADM):
 
     @staticmethod
     def kdma_value_to_system_prompt(kdma, value):
-        if kdma == "MoralDesert":
+        if kdma == "MoralDesert" or kdma=="Moral judgement":
             if value < 0.5:
                 return low_moral_deservingness_system_prompt()
             else:
@@ -133,46 +135,6 @@ class OutlinesTransformersADM(ActionBasedADM):
         prompt = action_selection_prompt(scenario_description, choices)
 
         return prompt, choices
-
-
-    def _format_single_incontext_prompt(self, sample, tgt_kdma, tgt_kdma_value):
-        """
-        Generate dialog for a single incontext example given the target kdma and value
-        """
-        state = sample["state"]
-        actions = sample["actions"]
-        labels = sample["labels"]
-
-        prompt, choices = self._state_to_top_level_prompt(state, actions)
-
-        # Compute each choice's distance from the target
-        # If the target attribute is not in the scores, assign infinite distance
-        # TODO: Add other metrics for determining "best" answer?
-        dist_to_tgt = [
-            abs(score[tgt_kdma] - tgt_kdma_value) if tgt_kdma in score else float('inf')
-            for score in labels
-        ]
-        # TODO: If there are ties, this always returns the first choice, do we need to chose differently?
-        correct_answer_idx = np.argmin(dist_to_tgt)
-
-        # TODO: Should we shuffle ICL options?
-
-        # TODO: Add chain of thought reasoning?
-        target_wording_mapping = {
-            "MoralDesert": "moral deservingness",
-            "maximization": "maximization"
-        }
-        adjective = "low" if tgt_kdma_value < 0.5 else "high"
-        reasoning_text = f"Per the principle of {adjective} {target_wording_mapping.get(tgt_kdma, tgt_kdma)}, " \
-                         f'\\"{choices[correct_answer_idx]}\\" is the correct answer.'
-
-        # TODO: Validate against schema?
-        answer = f'{{"detailed_reasoning": "{reasoning_text}", "action_choice": "{choices[correct_answer_idx]}"}}'
-
-        return [
-            {"role": "user", "content": prompt},
-            {"role": "assistant", "content": answer}
-        ]
 
     # Function borrowed from
     # https://docs.python.org/3/library/itertools.html#itertools.batched
@@ -203,8 +165,8 @@ class OutlinesTransformersADM(ActionBasedADM):
                                 alignment_target,
                                 num_positive_samples=1,
                                 num_negative_samples=0,
-                                shuffle_choices=False,
                                 generator_batch_size=5,
+                                kdma_descriptions_map='align_system/prompt_engineering/kdma_descriptions.yml',
                                 **kwargs):
         if self.baseline and num_negative_samples > 0:
             raise RuntimeError("No notion of negative samples for baseline run")
@@ -223,6 +185,7 @@ class OutlinesTransformersADM(ActionBasedADM):
 
         positive_icl_examples = []
         negative_icl_examples = []
+        incontext_settings=kwargs.get("incontext", {})
         if not self.baseline and alignment_target is not None:
             kdma_values = alignment_target.kdma_values
 
@@ -237,6 +200,10 @@ class OutlinesTransformersADM(ActionBasedADM):
             value = kdma_value['value']
             # Assumption here is that KDMA values range from 0-1
             negative_value = 1 - value
+            # Get kdma names and descriptions
+            with open(kdma_descriptions_map, 'r') as f:
+                kdma_descriptions = yaml.load(f, Loader=yaml.FullLoader)
+            name = kdma_descriptions[kdma]['name']
 
             positive_system_prompt = self.__class__.kdma_value_to_system_prompt(kdma, value)
             negative_system_prompt = self.__class__.kdma_value_to_system_prompt(kdma, negative_value)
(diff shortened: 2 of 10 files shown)


jadie1:
@dmjoy and @eveenhuis let me know what you think of this restructure. I left a TODO to add the KDMA value normalization option, I'll add that once we are happy with the format

align_system/utils/incontext_utils.py
```diff
@@ -0,0 +1,287 @@
...
+        Helper function for set_icl_datasets() - reads dataset files and gets examples for target_kdmas
+        Returns incontext_data dictionary with format:
+            {kdma:[{state, actions, choices, kdma_values}, ...], ...}
+        '''
+        incontext_data = {}
+        # For each kdma
+        for target_kdma in self.target_kdmas:
+            sys_kdma_name = target_kdma['kdma']
+            # Add examples for each dataset file
+            dset_files = self.incontext_settings["datasets"][sys_kdma_name]
```
eveenhuis, line 58: It's possible `sys_kdma_name` does not exist in `self.incontext_settings["datasets"]`, we either need to A) check and raise an exception appropriately or alternatively, B) just iterate over the `self.incontext_settings["datasets"]` dictionary.

B could result in loading more data than necessary, but it also might mean we could load the dataset just once at the beginning instead of every time we select an action (no matter the target, as long as we have something for each target KDMA we're given)

align_system/utils/incontext_utils.py
```diff
@@ -0,0 +1,287 @@
...
+            # If there is only one, make it a list for the following loop
+            if not isinstance(dset_files, list):
+                dset_files = [dset_files]
+            
+            # For each dataset file
+            for dset_f in dset_files:
+                with open(dset_f) as f:
+                    dset = json.load(f)
+                incontext_data[sys_kdma_name] = []
```
eveenhuis, line 67: If there are multiple dset files for one KDMA, I think this will overwrite with each new file. I **think** outside this for loop we'd want to do something like

```
if sys_kdma_name not in incontext_data:
    incontext_data[sys_kdma_name] = []
for dset_f in dset_files:
   ....
```
The if statement is even probably overkill

Apologies if that's not your intention with accepting a list of dset files per kdma now

align_system/utils/incontext_utils.py
```diff
@@ -0,0 +1,287 @@
...
+                        state,
+                        self.log
+                    )
+                    # Get KDMA_values
+                    kdma_values = []
+                    for label in icl_sample["label"]:
+                        if sys_kdma_name not in label:
+                            kdma_values.append(None)
+                        else:
+                            kdma_values.append(label[sys_kdma_name])
```
eveenhuis, line 85: If we ever get to datasets with multiple KDMAs per action, this will strip that information away. Our whole ICL strategy is one-KDMA-at-a-time focused right now, so I think this is fine, but wanted to call it out in case we think it might be useful to keep that information around

align_system/utils/incontext_utils.py
```diff
@@ -0,0 +1,287 @@
...
+        # Check that we have enough incontext examples for the target kdma
+        if len(possible_icl_examples) < n_icl_examples:
+            raise RuntimeError(f"Not enough possible incontext samples to learn from. Only "
+                            f"{len(possible_icl_examples)} samples available while asking for "
+                            f"{n_icl_examples} incontext samples.")
+        # If using LOO, don't include example ICL with exact same prompt
+        if self.incontext_settings.get("leave_one_out", False):
+            possible_icl_examples = [
+                icl_ex for icl_ex in possible_icl_examples
+                if icl_ex["prompt"] != prompt_to_match
```
eveenhuis, line 116: There is a use case where the prompt we check for LOO might be different than the prompt for similarity, e.g we may want to leave out the actions for LOO so that the ordering or a pre-filtering doesn't cause the matching to fail. I'm wondering if @dmjoy (or @jadie1!) has ideas for better matching strategies since this is pretty brittle anyway (yes, I know I originally wrote this :smile: )

align_system/utils/incontext_utils.py
```diff
@@ -0,0 +1,287 @@
...
+            ]
+
+        # Downselect to n_icl_examples via given method
+        icl_strategy = self.incontext_settings["method"]
+        
+        if icl_strategy == "random":
+            selected_icl_examples = random.sample(possible_icl_examples, n_icl_examples)
+        elif icl_strategy == "bert_similarity":
+            possible_icl_prompts = [icl_sample["prompt"] for icl_sample in possible_icl_examples]
```
eveenhuis, line 125: Similarly, there is a use case where the prompt we use for checking similarity might be slightly different than the prompt we want to present to the ADM in the end. The easy example for this is that maybe we only want to condition similarity on state description, not available actions. I do this in the [base Outlines ADM](https://github.com/ITM-Kitware/align-system/blob/main/align_system/algorithms/outlines_adm.py#L312-L317)

It might be a good check to try to implement that ICL method as well to make sure this interface is flexible enough

align_system/utils/incontext_utils.py
```diff
@@ -0,0 +1,287 @@
...
+class ComparativeRegressionIncontextExampleGenerator(IncontextExampleGenerator):
+    def __init__(self,
+                 incontext_settings, 
+                 target_kdmas, 
+                 log,
+                 **kwargs):
+        self.incontext_settings = incontext_settings
+        self.target_kdmas = target_kdmas
+        self.log = log
+        self.set_icl_datasets()
```
eveenhuis, line 148: I'm of the opinion that we shouldn't redefine this since it's doing the exact same thing as the base class, but open to hearing other opinions. At a minimum, I'm thinking it should be changed to `super().__init__()`?

align_system/utils/incontext_utils.py
```diff
@@ -0,0 +1,287 @@
...
+    
+    def set_icl_datasets(self):
+        icl_datasets = {}
+        incontext_data = self.read_icl_dataset_files()
+        
+        # Add each target to icl_datasets
+        for target_kdma in self.target_kdmas:
+            sys_kdma_name = target_kdma['kdma']
+            icl_datasets[sys_kdma_name] = []
+            kdma_incontext_data = incontext_data[sys_kdma_name]
```
eveenhuis, line 158: Might get a key error here. My earlier comment applies again

align_system/utils/incontext_utils.py
```diff
@@ -0,0 +1,287 @@
...
+                    # Only include choice if there is a ground truth KDMA value available
+                    if kdma_value is None:
+                        continue
+                    # Predicted scores are 0-10, KDMA values are 0-1
+                    scaled_kdma_value = int(kdma_value * 10)
+                    icl_response[choice] = {}
+                    icl_response[choice]['reasoning'] = self.get_chain_of_thought_reasoning(target_kdma, action,
+                                                                                            example['state'], choice,
+                                                                                            scaled_kdma_value)
+                    icl_response[choice]['score'] = scaled_kdma_value
```
eveenhuis, line 185: General comment, not expecting changes: I really wish that we could verify that this is consistent with the json schema, in case that changes

align_system/utils/incontext_utils.py
```diff
@@ -0,0 +1,287 @@
...
+                    icl_choices_with_outcomes[choice] = {'predicted_outcome':None}
+                icl_prompt = comparative_kdma_score_prediction_prompt(icl_scenario_description,
+                                                                    icl_choices_with_outcomes,
+                                                                    sys_kdma_name)
+                # Get example response
+                icl_response = {}
+                for action, choice, kdma_value in zip(example['actions'], example['choices'], example["kdma_values"]):
+                    # Only include choice if there is a ground truth KDMA value available
+                    if kdma_value is None:
+                        continue
```
eveenhuis, line 178: This skips non-labeled actions in the response schema, but they are still included in the prompt. This matches the DRE behavior, but we may want to considering pre-filtering before the prompt is generated

align_system/utils/incontext_utils.py
```diff
@@ -0,0 +1,287 @@
...
+
+                # Add example
+                icl_datasets[sys_kdma_name].append({
+                    "prompt": icl_prompt,
+                    "response": icl_response
+                    })
+                
+        self.icl_datasets = icl_datasets
+
+    def get_chain_of_thought_reasoning(self, target_kdma, action, state, choice, expected_value):
```
eveenhuis, line 195: It's possible that Chris' ICL method will pre-generate chain of thought reasoning. I'm not sure how that fits in here (or if it should). I don't know if that end up in a separate file (if so we'd need to match somehow), or in the original dataset, which means we'd need to extend the loading function. Maybe we're already going to have a split in "input-output" ICL datasets and "Chris" ICL datasets (I forgot what the library he's using is called)

For approaches where we don't predict KDMA values (like the base outlines ADM, even though that's not very valuable to us anymore), it's possible that we'd have pre-generated COT **per target**, so this could get really complicated

Additionally, I think the pre-generated COT is really supposed to be model-specific, so we'll need to come up with a better solution for managing these files if that happens

align_system/algorithms/outlines_adm.py
```diff
@@ -118,44 +119,15 @@ def kdma_value_to_system_prompt(kdma, value):
...
         """
         Generate prompt dialog based on given state and actions
         """
-        choices = self.format_choices(
+        choices = adm_utils.format_choices(
             [a.unstructured for a in actions],
             actions,
-            scenario_state
+            scenario_state,
+            log
```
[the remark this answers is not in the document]
jadie1: Oh nice, fixed here https://github.com/ITM-Kitware/align-system/pull/115/commits/b7f310394e0f1cf8651c683a4421567b38f6aaeb

align_system/utils/incontext_utils.py
```diff
@@ -0,0 +1,287 @@
...
+    '''
+    @abstractmethod
+    def __init__(self,
+                 incontext_settings, 
+                 target_kdmas, 
+                 log,
+                 **kwargs):
+        self.incontext_settings = incontext_settings
+        self.target_kdmas = target_kdmas
+        self.log = log
```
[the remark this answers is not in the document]
jadie1: Fixed here https://github.com/ITM-Kitware/align-system/commit/b7f310394e0f1cf8651c683a4421567b38f6aaeb

align_system/utils/incontext_utils.py
```diff
@@ -0,0 +1,287 @@
...
+    comparative_kdma_score_prediction_prompt
+)
+
+
+class IncontextExampleGenerator(object, metaclass=ABCMeta):
+    '''
+    Abstract class for incontext example generator
+    Instances of this class have unique set_icl_datasets() functions for formatting prompt and reponses 
+    '''
+    @abstractmethod
```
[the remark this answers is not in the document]
jadie1: Fixed here https://github.com/ITM-Kitware/align-system/pull/115/commits/48807cf3a519fba796786e9b845700dfae571dc6

align_system/utils/incontext_utils.py
```diff
@@ -0,0 +1,287 @@
...
+
+        The keys of self.icl_datasets are the 'kdma' keys from self.target_kdmas,
+        the values are a list of all ICL examples for that kdma,
+        each ICL example is a dictionary with keys: 'prompt', 'response'
+        For example: {kdma: [{'prompt':str, 'response':json}, ...], ...}
+        '''
+        self.icl_datasets = {}
+        pass
+
+    def read_icl_dataset_files(self):
```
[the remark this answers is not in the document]
jadie1: Fixed here: https://github.com/ITM-Kitware/align-system/pull/115/commits/48807cf3a519fba796786e9b845700dfae571dc6

align_system/utils/incontext_utils.py
jadie1: Good point, I went with option A here https://github.com/ITM-Kitware/align-system/pull/115/commits/48807cf3a519fba796786e9b845700dfae571dc6
When we initialize the generator we create the examples dictionary for all target KDMAs. So we are not loading more data than necessary but we also only load that dataset once, because then we can call `select_icl_examples()` with any of the target kdmas and it does not reload the dataset. Let me know if that makes sense or if I'm misunderstanding your point

align_system/utils/incontext_utils.py
jadie1: Good catch! Fixed here https://github.com/ITM-Kitware/align-system/pull/115/commits/48807cf3a519fba796786e9b845700dfae571dc6

align_system/utils/incontext_utils.py
jadie1: I didn't mean to copy `__init__` here, I agree we don't need it. Removed here https://github.com/ITM-Kitware/align-system/pull/115/commits/0ee831d9e81ad1dc08103214be8d0019dfe56da1

align_system/utils/incontext_utils.py
jadie1: I don't think think this will strip away the info in cases where there are multiple KDMAs per action. We loop over target kdmas one at a time and add examples for each target KDMA to the dictionary separately. So if an example does not contain one KDMA value we set that value to None, but if it contains another KDMA value for another of the target KDMAs it will still retain that and add the example there. We should retain all ground truth KDMA values for the target KDMAs

align_system/utils/incontext_utils.py
jadie1: That is a good point that the choice order or choice variations could cause this to fail. I guess we could just use the scenario description to check? That would require saving the scenario descriptions in addition to the prompts and responses in `self.icl_datasets` and then also passing the "scenario_description_to_match" in `select_icl_examples()`

align_system/utils/incontext_utils.py
jadie1: Good point, I will work on the base outlines ADM ICL implementation. I am thinking we should have two bert similarity strategy options - one for prompt matching and one for scenario description matching. That will require the change I described in the comment above so we can access the scenario descriptions

align_system/utils/incontext_utils.py
jadie1: I don't think we will get a key error because in  the `_read_icl_dataset_files()` function we do:
```
for target_kdma in self.target_kdmas:
    sys_kdma_name = target_kdma['kdma']
    incontext_data[sys_kdma_name] = []
```
There is a chance the list stays empty, but that will throw a runtime error in `select_icl_examples()`

align_system/utils/incontext_utils.py
jadie1: Good point, fixed here: https://github.com/ITM-Kitware/align-system/pull/115/commits/c5c7c851ff940c4dbee37272a390df8b7b1e000c

align_system/utils/incontext_utils.py
jadie1: Good idea, I added json schema validation here: https://github.com/ITM-Kitware/align-system/pull/115/commits/c5c7c851ff940c4dbee37272a390df8b7b1e000c

align_system/utils/incontext_utils.py
jadie1: I've added the two bert similarity options here https://github.com/ITM-Kitware/align-system/pull/115/commits/f4fdf9060b9ef0d5498e5ee691f060bdd0469f56
I'll reformat the base outlines ADM ICL next week

align_system/utils/incontext_utils.py
jadie1: I switched it to just use the scenario description here https://github.com/ITM-Kitware/align-system/commit/f4fdf9060b9ef0d5498e5ee691f060bdd0469f56
Let me know if this is better

align_system/utils/incontext_utils.py
jadie1: Yeah we will definitely have to give this more thought when we come to it. For Chris's approach we likely can't  use `_read_icl_dataset_files()`, we will need a new helper function for a different style of file that allows us to get target-specific reasoning

align_system/utils/incontext_utils.py
jadie1: I've reformatted the base outlines ADM ICL here https://github.com/ITM-Kitware/align-system/pull/115/commits/44cb70cb684cfb51c6bdd51faec3ed631a90dbee

jadie1:
@eveenhuis and @dmjoy I think I've addressed most of Emily's comments now. It's ready to be looked over again when you have a chance

dmjoy:
In generally this looks like an improvement (pulling most of the ICL code out into it's own module), and Emily provided some good low level feedback / suggestions.  I'm content to merge this as is, but have some high level thoughts maybe worth briefly discussing (so I can cover with a bit more nuance).

align_system/utils/incontext_utils.py
eveenhuis: > only load that dataset once

Right now, we construct a `IncontextExampleGenerator` per `top_level_choose_action` call, so we load the dataset every time. We could potentially move the incontext settings to the ADM instance settings and construct the ICL generator (load the dataset) once in `__init__`. But I think that'd mean we want to go with option B then, since (in theory, but not in reality yet), our KDMA alignment target could be different each time we choose an action

align_system/utils/incontext_utils.py
eveenhuis: I mentioned this on the call, but wanted to make it clearer: 

I think this is a good change, but it may not be general enough. If we keep strings in the dataset, we might want a "similarity_prompt" and a "loo_prompt" in addition to the final prompt in our dataset, that way it could be more general. In this particular function, I think that would be just renaming variables/keys, but would require slightly different dataset loading/generation

This is where I think @dmjoy's  comment about specifying functions for generating these prompts instead of strings would be really useful, but I can't picture what that would look like

Not something I would hold up this PR for though

align_system/utils/incontext_utils.py
eveenhuis: This comment is applicable if we load KDMAs based on incontext settings instead of target KDMAs

align_system/utils/incontext_utils.py
eveenhuis: See [related comment](https://github.com/ITM-Kitware/align-system/pull/115#discussion_r1742263054)

align_system/utils/incontext_utils.py
jadie1: Moving the ICL generator creation to the ADM `__init__` makes sense in the context of the comparative regression ADM and would be better. But for the base outlines ADM, the ICL example responses are dependent on the target KDMA values (because we need to know which choice is correct) and we don't have access to the target KDMAs in `__init__`. Comparative regression on the other hand doesn't use the target value to create responses,  only the ground truth KDMA values. I could move it for comparative regression only, but also the incontext settings are currently defined in the `inference_kwargs`, so I'd have to move that as well (which wouldn't work for the base outlines). 

@eveenhuis and @dmjoy I thought I could fix this without a larger restructure but now I don't think so. I'm inclined to leave it as is now and maybe open an issue to think about this more down the line. Thoughts?

dmjoy: Thanks for looking into it Jadie.  Yes I don't think it's worth more restructuring at this time and we can make a point to potentially revisit later on.  Unless @eveenhuis strongly objects, I'll go ahead and merge this in once the changelog has been update.

eveenhuis: I don't strongly object

CHANGELOG.md
```diff
@@ -5,8 +5,14 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## 0.5.2
```
dmjoy, line 6: @jadie1 quick fix here, could you create a new `Unreleased` section (rather than adding these changes under `0.5.2` which has already been tagged/released)

jadie1: Oops, yes I'll fix this

PR #117, opened 2024-09-05
ICL normalization corrections

jadie1:
I found a couple small mistakes through running the ICL normalization benchmark. This tiny PR corrects them

### DIFF
--- a/align_system/utils/incontext_utils.py
+++ b/align_system/utils/incontext_utils.py
@@ -92,13 +92,13 @@ class IncontextExampleGenerator(object, metaclass=ABCMeta):
             
             # Normalize ground truth KDMA values
             if 'normalization' in self.incontext_settings:
-                if self.incontext_settings['normalization'] != None or self.incontext_settings['normalization'] != 'rawscores':
+                if self.incontext_settings['normalization'] != None and self.incontext_settings['normalization'] != 'rawscores':
                     if self.incontext_settings['normalization'] == 'globalnorm':
                         incontext_data = self._global_normalization(incontext_data)
                     elif self.incontext_settings['normalization'] == 'localnorm':
                         incontext_data = self._local_normalization(incontext_data)
                     else:
-                        raise ValueError(f'"{self.normalization}" is not a valid incontext normalization option. '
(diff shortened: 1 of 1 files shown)


eveenhuis (approved):

align_system/utils/incontext_utils.py
```diff
@@ -92,13 +92,13 @@ def _read_icl_dataset_files(self):
...
             if 'normalization' in self.incontext_settings:
-                if self.incontext_settings['normalization'] != None or self.incontext_settings['normalization'] != 'rawscores':
+                if self.incontext_settings['normalization'] != None and self.incontext_settings['normalization'] != 'rawscores':
                     if self.incontext_settings['normalization'] == 'globalnorm':
                         incontext_data = self._global_normalization(incontext_data)
                     elif self.incontext_settings['normalization'] == 'localnorm':
                         incontext_data = self._local_normalization(incontext_data)
                     else:
-                        raise ValueError(f'"{self.normalization}" is not a valid incontext normalization option. '
+                        raise ValueError(f'{self.incontext_settings['normalization']} is not a valid incontext normalization option. '
```
eveenhuis, line 101: Missed this -- you need to use the other type of quotes here

eveenhuis (changes requested):

align_system/utils/incontext_utils.py
jadie1: You mean `f'{self.incontext_settings['normalization']}` -> `f'{self.incontext_settings["normalization"]}`

jadie1: I did just run it and the error printed correctly, but I can fix it

eveenhuis: Yes that's what I meant, I'm surprised that it worked as is

eveenhuis: Oh apparently this was a change in py3.12 -- since we support non-3.12 versions we need to use the "old" way

jadie1: Gotcha, fixed. Thanks Emily!

eveenhuis (approved):

PR #118, opened 2024-09-06
Outlines based personas ADM

dmjoy:
Hi @DavidMChan here's my initial integration of the personas provider with an outlines based ADM.  I've made one or two tweaks to the provider itself, the most significant of which is adding the option to filter out backstory probes based on what KDMAs are in the alignment target (on by default).  The thinking here is that we don't want to potentially confuse the LLM with other KDMAs we're not interested in with respect to the target, but maybe it's critical to the approach in some way to include everything?  Would love to hear your thoughts there.

One limitation to this initial version is that we don't have a way yet to select personas using a distribution based alignment target (i.e. a KDE), interested on your input there as well.  We've considered sampling from the target distribution and getting personas closest to the sampled value, but expect we need more than 3 personas to approximate the target, and as @eveenhuis pointed out this likely just approximates the maximum likelyhood value in the distribution since we're doing a majority vote of the persona responses.

In any case, I'm running a sweep over all alignment targets for the ADEPT training scenarios (Moral Judgement and Ingroup Bias) as the targets here are still single value; and hoping to see an improvement over the non-outlines scores we got before.  One additional note here, we're using the `mistralai/Mistral-7B-Instruct-v0.2` model (32k context window at least) as we're still working on getting Gemma to play nicely (will need to sync up with Emily about what the issue was there).

If you want to run this locally, I've created experiment configs for running against the TA3 server (namely `outlines_personas_adept_moral_judgement_train.yaml` and `outlines_personas_adept_ingroup_bias_train.yaml`), you can also run it against the example adept moral judgement file in the repo with: `run_align_system adm=outlines_persona align_to_target=true +alignment_target="ADEPT-DryRun-Moral judgement-1.0"` (doesn't produce an alignment score or anything but is a good smoke test).

### DIFF
--- a/align_system/algorithms/lib/persona/provider.py
+++ b/align_system/algorithms/lib/persona/provider.py
@@ -193,6 +193,8 @@ class PersonaProvider:
             backstory_sorter = ScalarAlignmentDistanceFunc(alignment_target_dict)
         elif target_type == AlignmentTargetType.KDE:
             backstory_sorter = KDEAlignmentDistanceFunc(alignment_target_dict)
+        else:
+            raise ValueError("ADM does not currently support a mix of scalar and KDE targets.")
 
         sorted_backstories = backstory_sorter(self._backstories)
 
--- a/align_system/algorithms/lib/persona/provider.py
+++ b/align_system/algorithms/lib/persona/provider.py
@@ -21,6 +21,7 @@ KDMA_TO_PROBE_MAPPING = {
     'Moral judgement': 'moral_judgment',
     'ingroup_bias': 'ingroup_bias',
     'IngroupBias': 'ingroup_bias',
+    'Ingroup Bias': 'ingroup_bias',
     'Ingroup bias': 'ingroup_bias',
     'value_of_life': 'value_of_life',
     'ValueOfLife': 'value_of_life',
--- a/align_system/algorithms/lib/persona/provider.py
+++ b/align_system/algorithms/lib/persona/provider.py
@@ -3,12 +3,16 @@ import logging
 import os
 import numpy as np
 import random
-from typing import List, Optional, Dict
+from typing import List, Optional, Dict, Callable
 
 from swagger_client.models import  AlignmentTarget
 
 from align_system.algorithms.lib.persona.types import Dialog, Backstory
 from align_system.algorithms.lib.persona.templates import BACKSTORY_ASSISTANT_PROMPT
+# 'personas_pe' imported here to be used as the base directory for the
+# 'backstories.json'; TODO: more consistent / long term plan for
+# storing database files etc. for algorithms
+from align_system.prompt_engineering import personas as personas_pe
 
 
 KDMA_TO_PROBE_MAPPING = {
@@ -32,11 +36,15 @@ KDMA_TO_PROBE_MAPPING = {
 logger = logging.getLogger(__name__)
 
 
+def _default_probe_filter(probe):
+    return True
+
+
 class PersonaProvider:
 
     def __init__(self,
                  backstory_collection: str = os.path.join(
-                    os.path.abspath(os.path.dirname(__file__)), "..", "prompt_engineering", "personas", "backstories.json"
+                    os.path.abspath(os.path.dirname(personas_pe.__file__)), "backstories.json"
                 ),
     ) -> None:
         """
@@ -137,8 +145,7 @@ class PersonaProvider:
         # Return the top N backstories
         return sampled_backstories
 
-
-    def _generate_dialog_from_backstory(self, backstory: Backstory) -> Dialog:
+    def _generate_dialog_from_backstory(self, backstory: Backstory,  probe_filter: Callable[[Dict], bool] = _default_probe_filter) -> Dialog:
         """
         Generates a dialog based on the provided backstory.
 
@@ -158,6 +165,9 @@ class PersonaProvider:
         }]
 
         for i, probe in enumerate(backstory['probes']):
+            if probe_filter is not None and not probe_filter(probe):
+                continue
+
             dialog.extend(
                 (
                     {"role": "user", "content": f'Question {i + 2}: {probe["probe_prompt"]}'}, # Question 2, 3, 4, ... etc.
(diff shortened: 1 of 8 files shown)


DavidMChan:
> select personas using a distribution based alignment target (i.e. a KDE), interested on your input there as well

I think that this is actually a really hard task - I've asked a couple of times what it means to make decisions on a distribution (since by nature, a decision is a point-estimate, and really, the only thing that you can do is pick the mean/center of one of the distributions) - we can maybe try to pick a panel which is likely to make decisions with the same distribution as the target distribution (using some kind of training set?) which may not need more than 3-5 personas, but that might require a separate optimization process for each distribution target. 

I was able to get Gemma working on my branch - do you know what the blocker is? Perhaps I've already solved it.

dmjoy:
Maybe I don't quite understand the nuance here, but my thinking was that yes we would pick the panel whose decisions would fall in the same distribution as the target.  (Implementation wise my thinking was that we would sample a single target value from the target distribution say 10 times or so, and for each of those sampled values, pick the closest backstory (that hasn't been used yet))

The blocker with Gemma seems to just be updating the transformers version (straightforward enough but we didn't want to start messing with dependencies right before the deadline; and I know that the latest version of transformers doesn't like how we're using the default chat templates under the covers, so requires some futzing there).  Can I ask what version of transformers you're running (and have been able to use with Gemma?); I can also go through their release notes and pick the `min(our version, version that supports Gemma)` and just see if that works.

dmjoy:
Also my experiment running over the ADEPT training data with this implementation (and filtering out the backstory probes that aren't part of the alignment target) didn't give great results (alignment scores are roughly on par with the non-outlines implementation).  I am re-running now without the backstory probe filtering in place to see if that helps, but not terribly hopeful.

DavidMChan:
> (Implementation wise my thinking was that we would sample a single target value from the target distribution say 10 times or so, and for each of those sampled values, pick the closest backstory (that hasn't been used yet))
I think I might not fully understand how this is supposed to work... In fact, now that I think about it, I'm not even entirely sure how the target distribution is specified. It might be worth setting up a quick meeting to talk through this. 

> Can I ask what version of transformers you're running
My transformers version is 4.42.3

>alignment scores are roughly on par with the non-outlines implementation
Hmm, that's interesting. Perhaps we need to look deeper into the logs to figure out why it's not altering the decisions as much. It would be good (if we want to continue to pursue this), to get some examples of where the alignment and the chosen samples are wildly different. It may be that the negative alignment is much worse (which might not be so surprising, given the hacks that had to be made to the standard ADM to get negative alignment working).

dmjoy:
Agreed that a meeting to discuss the distribution alignment target stuff makes sense.  Will ping you on Slack to get something on the calendar.

I upgraded transformers to 4.42.3 and am able to run the Gemma model now (namely `google/gemma-2-9b-it`, certainly let me know if there's a different flavor I should be using instead).  Will re-run my experiments with the Gemma model and circle back.

It does look like the alignment scores are lower on the more extreme ends of the alignment target spectrum (especially on the lower end) which indeed jives with the idea that it's more difficult to align to the lower end with this approach.  I'll try to gather some examples for our sync up.

align_system/algorithms/lib/persona/provider.py
```diff
@@ -32,11 +42,82 @@
...
+    def __call__(self, backstories):
+        # Now that we have the probe values, we can find a set of
+        # backstories that maximize the value.  For each backstory,
+        # add the value
+        backstories_with_values = []
+        for backstory in backstories:
+            value = sum(
+                np.abs(probe['response_value'] - self.probe_values[probe['probe']])
+                for probe in backstory['probes']
+                if probe['probe'] in self.probe_values
```
dmjoy, line 66: @DavidMChan just a heads up, I made this set of changes from the original as it was defaulting to a probe value of 0 if it wasn't in the alignment target (usually we only had the single KDMA of interest in there) so we were often just getting personas with the lowest probe values across the board (I think this was the bug we noticed in the call where the target was "Moral Judgement" of 0.0 and one of the three backstories had a value of "5").
```
-                np.abs(probe['response_value'] - self.probe_values.get(probe['probe'], 0))
+                np.abs(probe['response_value'] - self.probe_values[probe['probe']])
                 for probe in backstory['probes']

```

align_system/utils/alignment_utils.py
```diff
@@ -287,3 +294,25 @@ def _check_if_targets_are_kde(target_kdmas):
...
+            all_scalar_targets = False
+        if 'kdes' not in target_kdma or target_kdma["kdes"] is None:
+            all_kde_targets = False
+
+    if all_scalar_targets:
+        return AlignmentTargetType.SCALAR
+    elif all_kde_targets:
+        return AlignmentTargetType.KDE
+    else:
+        return AlignmentTargetType.MIXED
```
dmjoy, line 318: @jadie1 @eveenhuis Just FYSA, I've added this utility function here to replace this block of text we typically have in our ADMs

dmjoy:
@DavidMChan I needed to do a bit more refactoring than I wanted to do to get some version of distribution based persona selection working, so now there's an expectation that there's a callable that once initialized takes the backstories and produces a ranked list of matches (most matching to least matching).  Hopefully this is enough of an example about how to load / use the KDE targets to get you going, but let me know if not.

align_system/utils/alignment_utils.py
jadie1: Okay, looks good

align_system/algorithms/lib/persona/provider.py
```diff
@@ -102,32 +182,23 @@ def _choose_backstories_for_alignment_target(self, alignment_target: Optional[ty
...
         alignment_target_dict = alignment_target.to_dict() # type: ignore
-        probe_values = {
-            KDMA_TO_PROBE_MAPPING[k['kdma']]: k['value'] * 10
-            for k in alignment_target_dict.get('kdma_values', [])
-            if k['kdma'] in KDMA_TO_PROBE_MAPPING
-        }
 
         # Serialize the probe values into a repeatable string
-        probe_values_str = json.dumps(probe_values, sort_keys=True)
+        probe_values_str = yaml.dump(alignment_target_dict, sort_keys=True)
```
dmjoy, line 187: I needed to switch this to YAML as the KDE alignment targets aren't easily serializable to JSON

dmjoy:
@DavidMChan I'm planning to merge this PR in the next day or so (as it has at least one generally useful addition to the code base, and we're largely just adding functionality here).  Let me know if there are strong objections to merging this as-is (we can build onto it / fix things in another PR as needed)

DavidMChan:
No objections! Looks good as is - I've been working on tuning some things, but I don't want to be a blocker for overall progress.

PR #119, opened 2024-09-09
Add example config for outlines-based ADM determinism

eveenhuis:


eveenhuis:
@dmjoy ready for your review when you get a chance, really just combines your two example configs into one. Confirmed this works for the base Outlines ADM, comparative regression, and hybrid regression

dmjoy:
LGTM

PR #121, opened 2024-09-12
History-based cumulative KDE alignment option

jadie1:
Adds a new alignment function which uses the JS divergence between the target and a cumulative KDE (constructed using choice history) . This option is added to the oracle and comparative regression ADMs. To use, specify `distribution_matching: cumulative_kde`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 * Added a probabilistic option to alignment utilities. Exposed this option in oracle, comparative regression, and
   hybrid regression ADMs.
 * Example config for deterministic outlines-based ADM runs (`align_system/configs/experiment/examples/outlines_force_determinism.yaml`). Requires setting `force_determinsim` to true and using greedy sampler.
+* Added a history-based/cumulative KDE option to alignment utilities. Exposed this option in oracle and comparative regression.
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


dmjoy:
LGTM, will merge shortly unless there are any objections

eveenhuis:
I don't love that the new alignment function has a different return than the rest, but don't know if there's a better way to do it

dmjoy:
> I don't love that the new alignment function has a different return than the rest, but don't know if there's a better way to do it

Ah, I didn't catch that Emily.  Agreed that's not ideal, if it's easy to compute that outside of the alignment function (around where it's called to update the history) that's preferable (to keep a more consistent interface, and just to give that function less to do)

jadie1:
> > I don't love that the new alignment function has a different return than the rest, but don't know if there's a better way to do it
> 
> Ah, I didn't catch that Emily. Agreed that's not ideal, if it's easy to compute that outside of the alignment function (around where it's called, to update the history) that's preferable (to keep a more consistent interface, and just to give that function less to do)

Okay, I'll move it out of the alignment function and into the ADMs. Side note - the block of conditional code that calls the alignment function is getting large and it is copied and pasted into all ADMs. Maybe we should make a shared function

dmjoy:
Thanks Jadie, agreed that it's not very sustainable to have this large (and growing) conditional that needs duplication across several ADMs, I think it's alright to leave it as is for now or make a shared function.  Slightly longer term I do want to refactor the alignment function stuff to have a consistent interface and be instantiatable at the config level (similar to how the outlines sampler argument is now)

jadie1:
I've moved the history updating out of the alignment function

> Slightly longer term I do want to refactor the alignment function stuff to have a consistent interface and be instantiatable at the config level (similar to how the outlines sampler argument is now)

That makes sense, I will leave the conditional for now

PR #122, opened 2024-09-16
Choice info in input_output.json

jadie1:
This PR adjusts `choose_action()` to return a dictionary called `choice_info` in addition to the action to take. The `choice_info` dictionary contains ADM-specific info that would be difficult to parse from the log. `run_align_system.py` is updated to write `choice_info` to the `input_output.json` file for easier parsing. 

For comparative regression, the true and predicted KDMA values are logged and added to `choice_info`. This enables calculating the regression accuracy (MSE) from resulting `input_output.json` files, using this script: https://github.com/ITM-Kitware/scratch/blob/dev/regression_mse/calculate_mse.py

I added an empty dictionary for `choice_info` to the other ADMs for now. This could be updated down the line to include relevant info, such as the positive and negative votes in `outlines_adm.py`.

Open to suggestions about better ways of doing this, since it is not ideal to return arbitrary output from `choose_action()`.

### DIFF
--- a/align_system/algorithms/abstracts.py
+++ b/align_system/algorithms/abstracts.py
@@ -1,6 +1,6 @@
 from abc import ABC, abstractmethod
 
-from typing import Union
+from typing import Union, Tuple, Dict
 from swagger_client.models import State, Action, AlignmentTarget
 
 
@@ -10,7 +10,7 @@ class ActionBasedADM(ABC):
                       scenario_state: State,
                       available_actions: list[Action],
                       alignment_target: Union[type[AlignmentTarget], None],
-                      **kwargs) -> Action:
+                      **kwargs) -> Tuple[Action, Dict]:
         pass
 
 
--- a/align_system/algorithms/abstracts.py
+++ b/align_system/algorithms/abstracts.py
@@ -10,7 +10,7 @@ class ActionBasedADM(ABC):
                       scenario_state: State,
                       available_actions: list[Action],
                       alignment_target: Union[type[AlignmentTarget], None],
-                      **kwargs) -> Tuple[Action, Dict]:
+                      **kwargs) -> Union[Action, Tuple[Action, Dict]]:
         pass
 
 
--- a/align_system/algorithms/hybrid_kaleido_adm.py
+++ b/align_system/algorithms/hybrid_kaleido_adm.py
@@ -30,4 +30,5 @@ class HybridKaleidoADM(ActionBasedADM):
             action_to_take,
             dialog)
 
-        return action_to_take
+        choice_info = {}
+        return action_to_take, choice_info
--- a/align_system/algorithms/hybrid_kaleido_adm.py
+++ b/align_system/algorithms/hybrid_kaleido_adm.py
@@ -14,7 +14,7 @@ class HybridKaleidoADM(ActionBasedADM):
         self.outlines_adm = outlines_adm
 
     def choose_action(self, scenario_state, available_actions, alignment_target, **kwargs):
-        action_to_take = self.kaleido_adm.choose_action(
+        action_to_take, choice_info = self.kaleido_adm.choose_action(
             scenario_state, available_actions, alignment_target, **kwargs)
 
         # Build out initial dialog for the outlines ADM in order to
@@ -30,5 +30,4 @@ class HybridKaleidoADM(ActionBasedADM):
             action_to_take,
             dialog)
 
-        choice_info = {}
         return action_to_take, choice_info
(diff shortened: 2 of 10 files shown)


align_system/algorithms/hybrid_kaleido_adm.py
```diff
@@ -30,4 +30,5 @@ def choose_action(self, scenario_state, available_actions, alignment_target, **k
             action_to_take,
             dialog)
 
-        return action_to_take
+        choice_info = {}
```
eveenhuis, line 33: Nit: Not sure this needs to be changed now (and David would probably know better), but I'm guessing at a minimum the hybrid Kaleido ADM will pass along the dictionary from the non-hybrid Kaleido ADM

align_system/algorithms/outlines_regression_adm_comparative.py
```diff
@@ -300,6 +300,19 @@ def top_level_choose_action(self,
                                                         kdma_score_examples,
                                                         incontext_settings=kwargs.get("incontext", {}))
 
+        # Log true kdma values if present
+        true_kdma_values = {}
+        for choice_idx in range(len(available_actions)):
+            true_kdma_values[choices[choice_idx]] = available_actions[choice_idx].kdma_association
```
eveenhuis, line 306: Nit: Double checking that this doesn't cause issues for eval mode

eveenhuis (commented):
Overall LGTM, I don't have any strong opinions about this one

align_system/cli/run_align_system.py
```diff
@@ -272,7 +272,7 @@ def _compute_time_stats(times_s):
                 # prevent ADMs from modifying the originals (should
                 # considering doing the same for current_state and
                 # alignment_target)
-                action_to_take = adm.choose_action(
+                action_to_take, choice_info = adm.choose_action(
```
dmjoy, line 275: Would prefer to have this section of code work with either a single value or tuple return (to be more backward compatible and foolproof); would also need to update the type annotations here: https://github.com/ITM-Kitware/align-system/blob/main/align_system/algorithms/abstracts.py#L13

dmjoy: (FWIW even without supporting the different return shapes here, would want to update the type annotations)

align_system/algorithms/hybrid_kaleido_adm.py
jadie1: Ah good catch, fixed here https://github.com/ITM-Kitware/align-system/pull/122/commits/b38337170be9517731b7197833ce3f2e7efed6ec

align_system/algorithms/outlines_regression_adm_comparative.py
jadie1: I checked, it doesn't cause an issue, the true values are just logged as null. In the MSE calculation script, I made it so MSE is only calculated when the true values are available and not null.

align_system/cli/run_align_system.py
jadie1: Makes sense, I added the backward support here https://github.com/ITM-Kitware/align-system/pull/122/commits/44086be555f3110f439bcf106fdfd1705f37437b and updated the abstract type annotation here https://github.com/ITM-Kitware/align-system/pull/122/commits/6684d9e825aa15c1703500165d147d898ac9e37e

dmjoy: The updated code LGTM, thanks for making this change

align_system/algorithms/abstracts.py
```diff
@@ -10,7 +10,7 @@ def choose_action(self,
                       scenario_state: State,
                       available_actions: list[Action],
                       alignment_target: Union[type[AlignmentTarget], None],
-                      **kwargs) -> Action:
+                      **kwargs) -> Tuple[Action, Dict]:
```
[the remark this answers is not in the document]
jadie1: I believe that is right, I updated it here: https://github.com/ITM-Kitware/align-system/pull/122/commits/9b328d52f8b62309c27f71a8fed407efdc3ed3a9
What is the best way to test it? Calling `run_align_system` didn't throw an error when it was incorrect (`**kwargs) -> Action:` but `choice_info` was returned)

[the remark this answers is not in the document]
eveenhuis: Type hints are just promises, run a type checker like `ruff` or `mypy` to statically analyze your changes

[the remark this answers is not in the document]
dmjoy: FWIW I've been using the type hints as a form of documentation, but certainly worth running the static analysis tools to ensure they're all correct (but don't need to hold up the PR for this)

dmjoy:
I'm content to merge when you're ready @jadie1

jadie1:
@dmjoy I think it is good to go!

PR #123, opened 2024-09-17
Avoid potential divide by zero when misaligning

eveenhuis:
Removes possibility for divide by zero if all distances are zero when "misaligning." Also adds some additional error checking

### DIFF
--- a/align_system/utils/alignment_utils.py
+++ b/align_system/utils/alignment_utils.py
@@ -39,12 +39,18 @@ class AlignmentFunction(ABC):
         pass
 
     def _select_min_dist_choice(self, choices, dists, misaligned=False, probabilistic=False):
+        if len(dists) != len(choices):
+            raise RuntimeError("A distance must be provided for each choice during alignment")
+        if len(choices) == 0:
+            raise RuntimeError("No choices provided")
+
+        eps = 1e-16
         if not misaligned:
             # For aligned, want to minimize to distance to target
             # Invert distances so minimal distances have higher probability
 
             # Small epsilon for a perfect (0 distance) match
-            inv_dists = [1/(distance+1e-16) for distance in dists]
+            inv_dists = [1/(distance+eps) for distance in dists]
 
             # Convert inverse distances to probabilities
             probs = [inv_dist/sum(inv_dists) for inv_dist in inv_dists]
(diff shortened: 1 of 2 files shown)


dmjoy:
LGTM, no objections to merging.  Just a side note though; usually (if we're not dealing with numpy arrays here) I would explicitly check for a zero denominator and set the value to infinity or otherwise handle that condition more explicitly as using an epsilon can go unnoticed or be confusing if it isn't called out.

eveenhuis:
> LGTM, no objections to merging. Just a side note though; usually (if we're not dealing with numpy arrays here) I would explicitly check for a zero denominator and set the value to infinity or otherwise handle that condition more explicitly as using an epsilon can go unnoticed or be confusing if it isn't called out.

Your comment is making me question whether epsilon needs to be added to the numerator or not (and if that even matters). I was thinking that the values are only needed for relative comparison, so it probably (?) makes sense to operate on non-infinite values, but maybe not?

jadie1:
> Edit: I'm suggesting maybe it should just be ` probs = [dist/(sum(dists) + eps) for dist in dists]`

I think this would be sufficient

align_system/utils/alignment_utils.py
```diff
@@ -53,7 +59,9 @@ def _select_min_dist_choice(self, choices, dists, misaligned=False, probabilisti
             # maximize over non-inverted distances
 
             # Convert distances to probabilities
-            probs = [dist/sum(dists) for dist in dists]
+            # Avoid divide by zero if all distances are 0
+            total_dist = sum(dists) + eps * len(dists)
```
dmjoy, line 63: Does it make more sense here to do `sum(dists) * len(dists) + eps` (we don't need to multiply the epsilon right?)

dmjoy: (Or maybe I'm just confused about the intent here)

eveenhuis: Yeah, I think we should remove `len(dists)` from the equation entirely `sum(dists) + eps`

eveenhuis: Alright what do we think now? (`probs = [dist/(sum(dists) + eps) for dist in dists]`)

PR #124, opened 2024-09-19
Tweaks to log messages and input_output saving for easier parsing

dmjoy:
@aaron-bray PTAL, I think I've addressed each of your requests from the email thread.


aaron-bray:
Is it possible for the run_align_system to also pull and write out the ST KDMA target image to the results directory? 

Other than that, this has all the updates I requested

dmjoy:
You're talking about like a JPEG / PNG of the ST KDMA target?  That seems a bit out of scope for what `run_align_system` should do, but let me see what that would look like

aaron-bray:
Yeah, currently the target kdma json in the align_system.log looks like some sort of binary chunk of something...
Seems like the image is really the only thing that defines a ST target KDMA to a human, so it would be nice to have associated with the results to give them better context

aaron-bray:
This should be good, we can worry about ST KDE images in a differen PR

PR #126, opened 2024-09-26
Tweaks to input-output file and saved alignment target

dmjoy:
These changes are in support of the SoarTech alignment target(s) plotting script.

### DIFF
--- a/align_system/cli/run_align_system.py
+++ b/align_system/cli/run_align_system.py
@@ -2,7 +2,6 @@ import json
 from copy import deepcopy
 import atexit
 import os
-import yaml
 
 from rich.logging import RichHandler
 from rich.console import Console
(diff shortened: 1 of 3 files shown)


align_system/cli/run_align_system.py
```diff
@@ -144,10 +150,10 @@ def _compute_time_stats(times_s):
         else:
             log.info(alignment_target)
             if save_alignment_targets_to_path is not None:
-                alignment_target_path = os.path.join(save_alignment_targets_to_path, f"{alignment_target.id}.yaml")
+                alignment_target_path = os.path.join(save_alignment_targets_to_path, f"{alignment_target.id}.json")
 
                 with open(alignment_target_path, "w") as f:
-                    yaml.dump(alignment_target.to_dict(), f)
+                    json.dump(alignment_target.to_dict(), f, indent=2)
```
dmjoy, line 156: @eveenhuis just a heads up that I'm changing this output format to JSON (instead of YAML) to be consistent with the other output files we save from this script (and the above `OmegaConf.to_container` call is necessary to be able to serialize the target to JSON if we explicitly set the target with the Hydra config)

eveenhuis: Sounds good, we have [this script](https://github.com/ITM-Kitware/scratch/blob/master/get_soartech_alignment_targets.py) to pull down alignment targets if we ever need the yaml format anyway

PR #128, opened 2024-10-17
Adding prior norm for KDE alignment

jadie1:
This allows you to align to a KDE target normalized by the prior (frequency of values in train data).
To use set `adm.inference_kwargs.kde_norm` to `priornorm`
There are a couple TODOs:
- There is a normalization weighting parameter which is currently hard coded but should be added as a config param
- The frequency of values in the train data is currently hard coded - it should probably be calcuated from a specified training folder. I hard coded it by jointly counting the values across QoL and VoL, but there is a slight difference in the counts between KDMAs, so it may do better if we calculated a KDMA-specific prior.


align_system/utils/kde_utils.py
```diff
@@ -18,14 +18,38 @@ def load_kde(target_kdma, norm='globalnorm'):
...
+    elif norm == 'priornorm':
+        norm_factor = 0.2 #TODO - make this a config paramter
+
+        # Load target KDE and get density
+        kde = kde_from_base64(target_kdma['kdes']['rawscores']['kde'])
+        linspace = np.linspace(0, 1, 1000)
+        density = _kde_to_pdf(kde, linspace)
+
+        # Get prior KDE and density
+        prior_data = [0.1]*100 + [0.3]*12 + [0.7]*8 + [0.9]*100 #TODO - get from train yamls rather than hard code (counts*4)
```
dmjoy, line 35: Regarding getting the `prior_data` directly from the training data (rather than hardcoding); I believe the incontext database files could be used here (but I would want to pass in `prior_data` as a parameter here rather than a path to a training file or something to put less responsibility on this function).

That said, I don't think we need / want to worry about engineering this too much at this very moment since we already have long term refactoring planned for this part of the code, and we're coming up on the deadline soon.

dmjoy:
I don't have strong feelings one way or the other about these changes since they're additive in nature, and we're planning to refactor longer term.  I think whatever works for the Phase 1 eval works for me.

jadie1:
I've made the normalization weight a config param but the prior data distribution is still hard coded

align_system/algorithms/oracle_adm.py
```diff
@@ -21,7 +21,8 @@ def __init__(self, misaligned: bool=False, probabilistic: bool=False):
         self.choice_history = {} # Used for cumulative KDE alignment
 
     def choose_action(self, scenario_state, available_actions, alignment_target,
-                      distribution_matching='sample', kde_norm='rawscores', **kwargs):
+                      distribution_matching='sample', kde_norm='rawscores',
+                      norm_factor=0.5, **kwargs):
```
dmjoy, line 25: Just for my understanding, the `norm_factor` here only applies when the `kde_norm` value is `"priornorm"` is that right?  Just want to make sure that normalization isn't being applied by default.

align_system/utils/kde_utils.py
```diff
@@ -13,19 +13,47 @@
...
+        density = _kde_to_pdf(kde, linspace)
+
+        # Get prior KDE and density
+        #TODO - get prior data counts from train yamls rather than precomputing and hard coding
+        if target_kdma['kdma'] == 'QualityOfLife':
+            prior_data = [0.1]*24 + [0.3]*12 + [0.7]*12 + [0.9]*24
+        elif target_kdma['kdma'] == 'PerceivedQuantityOfLivesSaved':
+            prior_data = [0.1]*24 + [0.3]*16 + [0.7]*8 + [0.9]*24
+        else:
+            raise RuntimeError(f'No prior data for {target_kde['kdma']}')
```
dmjoy, line 39: This is a good error, so for example if we try to run prior norm on the ADEPT stuff it should error out loudly right?

dmjoy:
LGTM a few clarifying questions but otherwise happy to merge.

align_system/utils/kde_utils.py
jadie1: Yes running with ADEPT would trigger this error

align_system/algorithms/oracle_adm.py
jadie1: I'll change it quick

dmjoy: Thanks for making that change, I'll go ahead and merge in a few unless there's anything still outstanding from your end?

jadie1: I think it's good to go!

dmjoy:
Closing as these changes have been merged in with #138

PR #130, opened 2024-11-05
Adding option to sort ICL examples

jadie1:
Adds option to sort ICL examples responses (`sort_actions` in the `incontext` settings). 
This enables matching the ICL and probe response order when `sort_available_actions: true`

### DIFF
--- a/align_system/configs/adm/outlines_regression_aligned_comparative/incontext.yaml
+++ b/align_system/configs/adm/outlines_regression_aligned_comparative/incontext.yaml
@@ -7,8 +7,11 @@ inference_kwargs:
     number: 5
     method: prompt_bert_similarity
     leave_one_out: false
-    normalization: null
-    sort_actions: true
+    normalization: globalnorm
     datasets:
-      QualityOfLife: /data/shared/samba/phase1_icl/qol_train1_20241105.json
-      PerceivedQuantityOfLivesSaved: /data/shared/samba/phase1_icl/vol_train1_20241105.json
+      MoralDesert: /data/shared/samba/integrated_results_metrics_eval/captured_dataset_for_chris/baseline_adept_high-1715105775-input-output.json
+      maximization: /data/shared/samba/integrated_results_metrics_eval/captured_dataset_for_chris/baseline_soartech_high-1716581856-input-output.json
+      Moral judgement: /data/shared/samba/dry_run/moral_judgement_20240826.json
+      Ingroup Bias: /data/shared/samba/dry_run/ingroup_bias_20240826.json
+      QualityOfLife: /data/shared/samba/dry_run/qol_20240826.json
+      PerceivedQuantityOfLivesSaved: /data/shared/samba/dry_run/vol_20240826.json
--- a/align_system/configs/adm/outlines_regression_aligned_comparative/incontext.yaml
+++ b/align_system/configs/adm/outlines_regression_aligned_comparative/incontext.yaml
@@ -7,11 +7,8 @@ inference_kwargs:
     number: 5
     method: prompt_bert_similarity
     leave_one_out: false
-    normalization: globalnorm
+    normalization: null
+    sort_actions: true
     datasets:
-      MoralDesert: /data/shared/samba/integrated_results_metrics_eval/captured_dataset_for_chris/baseline_adept_high-1715105775-input-output.json
-      maximization: /data/shared/samba/integrated_results_metrics_eval/captured_dataset_for_chris/baseline_soartech_high-1716581856-input-output.json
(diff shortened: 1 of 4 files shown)


eveenhuis (approved):

align_system/configs/adm/outlines_regression_aligned_comparative/incontext.yaml
```diff
@@ -7,11 +7,8 @@ inference_kwargs:
     number: 5
     method: prompt_bert_similarity
     leave_one_out: false
-    normalization: globalnorm
+    normalization: null
+    sort_actions: true
```
dmjoy, line 11: I have some concern about changing default options without more fanfare, as someone could think they're re-running a previous config that we submitted for evaluation but be using these new settings instead without really noticing (something I probably would have done).  Does this actually change the `normalization` or is the default behavior (if `normalization: null` to use `globalnorm`)?

align_system/configs/adm/outlines_regression_aligned_comparative/incontext.yaml
```diff
@@ -7,11 +7,8 @@ inference_kwargs:
...
     datasets:
-      MoralDesert: /data/shared/samba/integrated_results_metrics_eval/captured_dataset_for_chris/baseline_adept_high-1715105775-input-output.json
-      maximization: /data/shared/samba/integrated_results_metrics_eval/captured_dataset_for_chris/baseline_soartech_high-1716581856-input-output.json
-      Moral judgement: /data/shared/samba/dry_run/moral_judgement_20240826.json
-      Ingroup Bias: /data/shared/samba/dry_run/ingroup_bias_20240826.json
-      QualityOfLife: /data/shared/samba/dry_run/qol_20240826.json
-      PerceivedQuantityOfLivesSaved: /data/shared/samba/dry_run/vol_20240826.json
+      QualityOfLife: /data/shared/samba/phase1_icl/qol_train1_20241105.json
+      PerceivedQuantityOfLivesSaved: /data/shared/samba/phase1_icl/vol_train1_20241105.json
```
dmjoy, line 14: Wondering why we're removing the other KDMAs here, are we only using incontext for the SoarTech KDMAs now?  (Maybe I'm missing some context for these changes)

align_system/configs/adm/outlines_regression_aligned_comparative/incontext.yaml
jadie1: When I added the option to normalize ICL scores I changed this default and shouldn't have. null is equivalent to rawscores or no normalization which was the default for DRE. I think it makes more sense to turn this off by default, I don't think we will end up using ICL norm in the final config

align_system/configs/adm/outlines_regression_aligned_comparative/incontext.yaml
jadie1: We haven't created the updated Phase1 ICL files for Adept yet. I took them out for now so that we didn't have a mix of old (DRE) and new ICL jsons in there

jadie1: These SoarTech files will also eventually be replaced with ones created from the final versions of all three train files

align_system/configs/adm/outlines_regression_aligned_comparative/incontext.yaml
dmjoy: Ah gotcha, good to know.  I don't have much objection here then (and I suspect that changing the `sort_actions` to be true by default doesn't make a tremendous impact on system behavior?).  When you update the changelog please note the change in the default behavior in the `Changed` section so there's a more noticeable record of the change

align_system/configs/adm/outlines_regression_aligned_comparative/incontext.yaml
dmjoy: So I think the problem I'm still having then is that if we merge this in as-is, we won't be able to smoketest ICL on adept; I feel like a smoother transition would be to just update the ICL databases that are updated.  Would a feasible alternative here be to create a new config file here called `incontext_phase1.yaml` or something like that, which can be more volatile leading up the the eval, but the `incontext.yaml` is the stable older version used for DRE (or maybe `incontext_dre.yaml`?

jadie1: Makes sense, I'll change it to `incontext_phase1.yaml` and keep the old one

jadie1: Done

dmjoy: Thanks Jadie, merging shortly

PR #132, opened 2024-11-06
Add character-based LOO

eveenhuis:
Adds character based LOO. Removes duplicate ICL examples, based on configured similarity measure. Phase 1 experiment directory created.

Note:
- `incontext.leave_one_out=false`  should now be specified as `incontext.leave_one_out=null`
- `incontext.leave_one_out=true`  should now be specified as `incontext.leave_one_out=scenario_description` (old behavior) or `incontext.leave_one_out=characters`

### DIFF
--- a/align_system/algorithms/oracle_adm.py
+++ b/align_system/algorithms/oracle_adm.py
@@ -34,9 +34,9 @@ class OracleADM(ActionBasedADM):
         all_scalar_targets = True
         all_kde_targets = True
         for target_kdma in target_kdmas:
-            if not hasattr(target_kdma, 'value') or target_kdma.value is None:
+            if 'value' not in target_kdma or target_kdma['value'] is None:
                 all_scalar_targets = False
-            if not hasattr(target_kdma, 'kdes') or target_kdma.kdes is None:
+            if 'kdes' not in target_kdma or target_kdma['kdes'] is None:
                 all_kde_targets = False
 
         # get ground truth kdma values
--- a/align_system/configs/experiment/phase1_evaluation/random_adept_ingroup_bias_train.yaml
+++ b/align_system/configs/experiment/phase1_evaluation/random_adept_ingroup_bias_train.yaml
@@ -1,0 +1,18 @@
+# @package _global_
+defaults:
+  - override /adm: random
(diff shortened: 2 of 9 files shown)


dmjoy (approved):
Seems fine to me, minor preference for making the parameter `incontext.leave_one_out_strategy` as the current name strongly implies a boolean (that and changing the parameter name will hopefully break more loudly and obviously if someone tries `incontext.leave_one_out=true` etc. moving forward since that parameter won't exist)

align_system/algorithms/oracle_adm.py
jadie1, line 1: @dmjoy This is the same fix I've been using to handle the targets in dictionary format

align_system/configs/experiment/phase1_evaluation/random_adept_ingroup_bias_train.yaml
jadie1, line 1: We may want to add `force_determinism` for the random runs

jadie1 (approved):
Looks good to me!

align_system/configs/experiment/phase1_evaluation/random_adept_ingroup_bias_train.yaml
eveenhuis: Then they aren't random :) I tend to use the random experiments as a base for my other experiments (so I wouldn't prefer the change), but I can make the change if we think it makes sense

eveenhuis:
> Seems fine to me, minor preference for making the parameter `incontext.leave_one_out_strategy` as the current name strongly implies a boolean (that and changing the parameter name will hopefully break more loudly and obviously if someone tries `incontext.leave_one_out=true` etc. moving forward since that parameter won't exist)

@dmjoy done!

dmjoy:
Thanks @eveenhuis, merging in just a few

align_system/configs/experiment/phase1_evaluation/random_adept_ingroup_bias_train.yaml
jadie1: I'm fine to leave it then. I guess if we wanted to truly benchmark random performance we'd have to average across multiple runs without determinism

PR #135, opened 2024-11-06
Adding history-based scalar target alignment 

jadie1:
Adds history-based (running mean) alignment function for scalar targets.  Added only to oracle ADM for now.


dmjoy:
LGTM

PR #136, opened 2024-11-07
Enumerate valid scores in json schema

jadie1:
Added the option to enumerate the valid regression scores in the json schema by specifying 
`inference_kwargs.enum_scores` as true. Valid score options for each KDMA are added to `align_system/prompt_engineering/kdma_descriptions.yml` in a new field: `valid_scores`.


dmjoy:
LGTM!

PR #140, opened 2024-11-11
Add KDMA scale factor for comparative regression

eveenhuis:
Add option to scale KDMA values (e.g 0-100 instead of 0-10). Only implemented for comparative regression ADM (not non-comparative regression ADM)

Also updates the enum values option to accept either a list or range parameters, so that a user has more options to specify values as the possible values grow in number


jadie1:
@eveenhuis Just to make sure I am understanding correctly, the scale factor is always used regardless of whether `enum_scores` is true or false. So if you set the scale factor to 10 this would be equivalent to the current main branch, is that correct?

eveenhuis:
> @eveenhuis Just to make sure I am understanding correctly, the scale factor is always used regardless of whether `enum_scores` is true or false. So if you set the scale factor to 10 this would be equivalent to the current main branch, is that correct?

Yep!

jadie1 (approved):
I think this looks good. We are starting to put things in `kdma_descriptions.yml` that feel like they should maybe be in the config. But I like that we can potentially set them to different values for different KDMAs and I think it is okay for Phase 1. We just need to double check both `kdma_descriptions.yml` and the final config before the submission

eveenhuis:
> I think this looks good. We are starting to put things in `kdma_descriptions.yml` that feel like they should maybe be in the config. But I like that we can potentially set them to different values for different KDMAs and I think it is okay for Phase 1. We just need to double check both `kdma_descriptions.yml` and the final config before the submission

Yeah, I agree. I really wish the `factor` especially was in the config, but I felt like I couldn't separate it from the `score_examples` since they directly utilize it. Maybe you feel differently, though!

jadie1:
> Yeah, I agree. I really wish the `factor` especially was in the config, but I felt like I couldn't separate it from the `score_examples` since they directly utilize it. Maybe you feel differently, though!

Yeah I don't see an obvious way to do it. It's good to merge as is from my perspective

align_system/prompt_engineering/kdma_descriptions.yml
```diff
@@ -1,35 +1,52 @@
...
+  factor: 10
+  score_examples: 'Consider a situation where patient A and patient B are both injured. If both patients acted in a morally neutral manner, choosing to treat patient A would receive a score of {{ "%g" | format(0.5 * kdma_scale_factor) }} for moral deservingness. However if patient A caused harm or acted with poor morality, choosing to treat patient A would receive a score of {{ "%g" | format(0 * kdma_scale_factor) }}, while choosing to treat patient B would receive a score of {{ "%g" | format(1.0 * kdma_scale_factor) }}.'
```
dmjoy, line 5: I'm assuming the `"%g"` here formats into an integer?  (It's tough to google such a thing)

eveenhuis: It's "standard number format" or something along those lines, but yes 10.0 becomes 10, but 10.5 would stay 10.5

PR #151, opened 2024-11-14
Action type based ICL selection

jadie1:
Added ICL example selection method that gives larger weight to examples with the same action types as the current probe. To use set `incontext.method` to `matching_actions`.

If there are no ICL examples with the same action types then `prompt_bert_similarity` is used.


dmjoy:
LGTM, merging shortly

PR #155, opened 2024-12-06
Add multi-KDMA targets for testing

eveenhuis:
Adds multi-kdma targets for testing

`Ingroup Bias-lowhigh` means IB is low and MJ is high whereas `Moral judgement-lowhigh` means MJ is low and IB is high, i.e the KDMA that will be scored by the TA1/3 server is listed first


jadie1:
The targets look good, I think the names are a little confusing. Could we put both the KDMAs in them? Maybe something like `IO-low_MJ-high`?

jadie1:
Since these are not the "final" targets from TA1s I don't think the naming is super important if we just want to get this merged

eveenhuis:
Target files have been renamed

dmjoy (approved):

PR #156, opened 2024-12-18
Relevance

jadie1:
This PR adds:
- Unofficial Multi-KDMA distribution alignment targets
- Multi-KDMA target alignment functions that weight distance by relevance
- An oracle ADM that uses relevance
- Option to use Kaleido relevance with scalar targets
- Option to for comparative regression ADM to predict relevance 
- Relevance ICL for comparative regression


dmjoy:
No objections to merging from my side.

eveenhuis:
LGTM!

PR #158, opened 2024-12-19
Updating Kaleido Alignment Functions

jadie1:
Adds the option to use cumulative KDE alignment with Kaleido, either with or without relevance.
- This config runs cumulative KDE alignment without relevance: `align_system/configs/experiment/phase1_evaluation/hybrid_kaleido_soartech_train.yaml`
- This config runs with relevance: `align_system/configs/experiment/phase1_evaluation/relevance_hybrid_kaleido_soartech_train.yaml`

### DIFF
--- a/align_system/utils/alignment_utils.py
+++ b/align_system/utils/alignment_utils.py
@@ -137,7 +137,7 @@ class AvgDistScalarAlignment(AlignmentFunction):
 
 
 class CumulativeAvgDistScalarAlignment(AlignmentFunction):
-    def __call__(self, kdma_values, target_kdmas, choice_history, misaligned=False, kde_norm='globalnorm', probabilistic=False):
+    def __call__(self, kdma_values, target_kdmas, choice_history={}, misaligned=False, kde_norm='globalnorm', probabilistic=False):
         '''
         Uses choice history to calcualte a running average,
         selects the choice that brings the running average closeest to the scalar target.
--- a/align_system/utils/alignment_utils.py
+++ b/align_system/utils/alignment_utils.py
@@ -137,13 +137,16 @@ class AvgDistScalarAlignment(AlignmentFunction):
 
 
 class CumulativeAvgDistScalarAlignment(AlignmentFunction):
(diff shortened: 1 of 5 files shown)


align_system/utils/alignment_utils.py
```diff
@@ -137,7 +137,7 @@ def get_best_sample_index(self, kdma_values, target_kdmas, selected_choice, misa
 
 
 class CumulativeAvgDistScalarAlignment(AlignmentFunction):
-    def __call__(self, kdma_values, target_kdmas, choice_history, misaligned=False, kde_norm='globalnorm', probabilistic=False):
+    def __call__(self, kdma_values, target_kdmas, choice_history={}, misaligned=False, kde_norm='globalnorm', probabilistic=False):
```
eveenhuis, line 140: Nit: I'm not sure it makes sense to have a default arg for the choice history? 

I'm not seeing where you make use of it, though

eveenhuis: Also, side note: apparently mutable objects as default args is bad practice because they are only initialized once. If we *do* want a default value here, the "better-practice" way to do so would be `choice_history=None` and then in the function body 

```
if choice_history is None
    choice_history = {}
```

so that we don't end up in any weird states

jadie1: I had to add this because of this weird way we have to call the alignment function in kaleido: https://github.com/ITM-Kitware/align-system/blob/6105d504967201d2b508a324d6c7e3d9e8953a00/align_system/algorithms/kaleido_adm.py#L374
So we can call it in a consistent way here:
https://github.com/ITM-Kitware/align-system/blob/6105d504967201d2b508a324d6c7e3d9e8953a00/align_system/algorithms/kaleido_adm.py#L415
@dmjoy Is there a smoother way to do this? 
I can at least change it to `None` as @eveenhuis suggested

dmjoy: I thinking changing the default to `None` makes sense; the `partial` stuff is more of a band-aid around the alignment functions not having a common interface (something I hope to fix with some longer term refactoring).

dmjoy (approved):

PR #159, opened 2024-12-19
Log true relevance for comp regression

eveenhuis:
Adds true relevance values to the input output file for the comparative regression ADM. I don't think this is worthy of a changelog entry, but let me know if you'd like me to add one


dmjoy:
LGTM, so this just takes the presence of the KDMA as a relevance value of 1 right

dmjoy (approved):

eveenhuis:
> LGTM, so this just takes the presence of the KDMA as a relevance value of 1 right

Yes, that's what the accuracy script does as well. I don't think this is a critical change since I don't actually use the field in the accuracy script. However, if we keep it in Jadie's MR, we should probably merge this in for consistency.

jadie1 (approved):

PR #161, opened 2025-01-02
Tweaks for multi-KDMA experiment runs

dmjoy:
Please take a look @jadie1 @eveenhuis and make sure I haven't made a mistake with the configs.  If you all run more experiments etc. feel free to update this branch directly, or I can merge it and you all can make follow-up MRs.


eveenhuis (commented):
I think these all look right to me

jadie1 (approved):
These all look correct to me

PR #162, opened 2025-01-30
Dev/add integration tests

dmjoy:
See documentation for how to run / add tests here: https://github.com/ITM-Kitware/align-system/blob/390983775440901b5708bd2113faf4076595c58b/docs/integration_testing.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 * Added a relevance oracle ADM
 * Added an optional relevance prediction step with ICL to the comparative regression ADM
 * Added options to use cumulative KDE and/or relevance weighted alignment function to Kaleido ADM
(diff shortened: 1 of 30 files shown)


dmjoy:
@eveenhuis @jadie1 when you all get a chance, please take a look at this MR and suggest (or add) any configs you thing are important to put under test.  This functionality could also be useful to you as you update / modify your own ADMs.

eveenhuis (approved):
LGTM. Only things to note that I don't think are critical before this first merge:

- We're obviously only testing comparative regression outputs as is, but maybe that's all we care about right now. There's pros & cons to being more or less exhaustive so I'll leave that decision up to you
- It' might be a good idea to add a multi-KDMA test for functionality reasons, I think the output here is less important since it's still very much under development, but we don't want to break that actual functionality

jadie1 (approved):
Looks good I think you've added the most important configs. I agree with Emily we may want to add one with relevance and maybe also Kaliedo, but that can be done later. No objections to merging now

PR #163, opened 2025-01-30
Single heuristic treatment bugfix

eveenhuis:
Action parameters weren't being filled in with a single heuristic treatment option, so we were falling back to the "old" way of choosing treatments.


dmjoy:
LGTM, merging

PR #164, opened 2025-02-25
Explore pipeline based ADM refactorization

dmjoy:
(NOT INTENDED TO BE MERGED IN CURRENT STATE)

This is an exploratory PR for a pipeline style ADM approach.  More details forthcoming


align_system/algorithms/abstracts.py
```diff
@@ -1,8 +1,10 @@
 from abc import ABC, abstractmethod
 
-from typing import Union, Tuple, Dict
+from typing import Union, Tuple, Dict, List
```
eveenhuis, line 3: *very* minor, but since we are 3.9+, you don't need to import some of these and can just use the lowercase version in your type hints e.g `tuple`, `dict`, `list`. Looks like you actually have some mixing of lower vs upper in this file

align_system/algorithms/alignment_adm_component.py
```diff
@@ -0,0 +1,48 @@
...
+                 alignment_function,
+                 attributes={}):
+        self.alignment_function = alignment_function
+        self.attributes = attributes
+
+    def run(self,
+            scenario_state,
+            choice_evaluation,
+            dialogs,
+            alignment_target=None):
```
eveenhuis, line 16: Nit: If we're going to spend the time refactoring to make things more usable/readable, I find it's easier to understand what's happening if type hints are also included in impls, but that might just be me. This would also set us up better if we ever get to the point where we want to turn on type checking

align_system/algorithms/alignment_adm_component.py
```diff
@@ -0,0 +1,48 @@
...
+        target_kdmas = alignment_target_to_attribute_targets(
+            alignment_target,
+            self.attributes)
+
+        # Alignment function below is expecting target_kdmas to be dicts
+        target_kdmas = [dict(t) for t in target_kdmas]
+
+        predicted_kdma_values =\
+            {k: v['kdma_prediction_scores']
+             for k, v in choice_evaluation.items()}
```
eveenhuis, line 30: Seems very geared towards regression ADMs

align_system/algorithms/abstracts.py
```diff
@@ -14,6 +16,28 @@ def choose_action(self,
...
+        pass
+
+
+class ADMComponent(ABC):
+    @abstractmethod
+    def run(self,
+            scenario_state: State,
+            choice_evaluation: Dict,
+            dialogs: List[Dialog],
+            alignment_target: Union[type[AlignmentTarget], None]) -> Tuple[Dict, List[Dialog]]:
```
eveenhuis, line 36: Slightly weird that we're going to be strict on inputs but have no expectations on outputs, but I get that you want to keep it as flexible as possible

align_system/algorithms/pipeline_adm.py
```diff
@@ -0,0 +1,60 @@
+from align_system.algorithms.abstracts import ActionBasedADM
+from align_system.utils import adm_utils
+from align_system.utils import logging
+
+log = logging.getLogger(__name__)
+
+
+class PipelineADM(ActionBasedADM):
+    def __init__(self,
+                 steps={}):
```
eveenhuis, line 10: Having a mutable default argument is a no-no in python as it only gets initialized once. The safer way to do this is setting default to `None` and then initializing the function body e.g `if steps is None`... though not entirely sure what the point of an empty pipeline would be. Also, should we type hint this since you're describing the type in a comment?

eveenhuis (commented):
Just wanted to get some of my pre-meeting thoughts written down, possible that my opinions change after the meeting :)

align_system/algorithms/alignment_adm_component.py
dmjoy: Yes that's very true, my assumption is that for ADMs that "just choose" and answer would just set the choice_evaluation key `"chosen": True` for the answer it picks and wouldn't even hit alignment part of the code.  To make that more obvious this class could be called the `RegressionAlignmentADMComponent` (?)

align_system/algorithms/abstracts.py
dmjoy: Aye, just trying to be a bit more strict where we can be

align_system/algorithms/pipeline_adm.py
dmjoy: That's right, I think you mentioned that before also, I'll make that change.

align_system/algorithms/alignment_adm_component.py
eveenhuis: I think that makes sense

dmjoy:
Somewhat major change to how the pipeline steps are invoked, and data shared between them.  Each step is required to specify what their `run` method return (via `run_returns`), and these outputs are added into a `working_output` dictionary.  When the next step is invoked, there's a new utility function `call_with_coerced_args` that takes the step function and a dictionary, and using the `inspect` module figures out how to populate the parameters for the call with what's in the `working_output` (complaining if it doesn't have everything it needs).

dmjoy:
Planned TODO:
- [ ] Update current pipeline components to also call prompts/templates with the `call_with_coerced_args` function (where the ADM component developer should build a dictionary of things they expect the prompts to possibly need, and prompts (functions) only need to specify what parameters they actually intend to use, rather than imposing the same (almost certainly) superset of arguments that each prompt should have).
- [ ] Create a new driver script for pipeline ADMs (similar to `run_align_system`) that doesn't do any ITM specific things, and create ADM components to handle the ITM stuff (perhaps this requires a rethinking of the input-output format and/or acceptance that it might look slightly different based on the domain)

align_system/algorithms/alignment_adm_component.py
```diff
@@ -0,0 +1,39 @@
...
+        target_kdmas = alignment_target_to_attribute_targets(
+            alignment_target,
+            self.attributes)
+
+        # Alignment function below is expecting target_kdmas to be dicts
+        target_kdmas = [dict(t) for t in target_kdmas]
+
+        selected_choice, probs = self.alignment_function(
+            attribute_prediction_scores,
+            target_kdmas)
```
dmjoy, line 32: Should change this to use `call_with_coerced_args`

align_system/configs/adm/pipeline.yaml
```diff
@@ -0,0 +1,52 @@
...
+  - /inference_engine@structured_inference_engine: outlines_structured_inference_engine
+  - /template/scenario_description@scenario_description_template: with_relevant_char_info
+  - /template/prompt@prompt_template: comparative_regression
+  - /template/output_schema@comparative_regression_choice_schema: comparative_regression_choice
+  # ADM components to be used in "steps"
+  - /adm_component/misc@step_definitions.format_choices: itm_format_choices
+  - /adm_component/icl@step_definitions.regression_icl: regression
+  - /adm_component/regression@step_definitions.comparative_regression: comparative
+  - /adm_component/alignment@step_definitions.scalar_alignment: avg_dist_scalar
+  - /adm_component/misc@step_definitions.justification_from_reasonings: justification_from_reasonings
```
dmjoy, line 18: @jadie1 note the additional step here to get the justification for the selected action

align_system/configs/adm_component/regression/comparative.yaml
```diff
@@ -0,0 +1,7 @@
+_target_: align_system.algorithms.comparative_regression_adm_component.ComparativeRegressionADMComponent
+
+structured_inference_engine: ${ref:adm.structured_inference_engine}
+num_samples: 2
```
jadie1, line 4: If greedy sampling is the default we should make the default num_samples 1

dmjoy: Good catch Jadie (I'll make that change), and apologies I should have mentioned making the default the greedy sampler as I know that's something we discussed but didn't commit to and probably has some impact on your latest experiment setup.

CHANGELOG.md
```diff
@@ -3,14 +3,20 @@
...
 
 * Fixed non-determinism issues with outlines_adm baseline with shuffle_choices arg and outlines seed
+* Fixed configs referenced in top-level README for Phase 1 Evaluation runs
 
 ### Added
 
+* Added PipelineADM and many pipeline ADM components, configs, and integration tests
+* Added new data models for Attributes and Dialogs
+* Added `call_with_coerced_args` utility function to handle calling functions with different input requirements
+* Added documentation for new pipeline ADM developement
```
jadie1, line 18: Would it make sense to link the documentation here?

docs/pipeline_adms.md
```diff
@@ -0,0 +1,221 @@
...
+algorithmic code, as we already have several Pipeline ADM steps that
+you can re-use to handle much of the data wrangling that may be
+necessary.  In this document we'll cover what constitutes a pipeline
+ADM, how to create pipeline ADM components (steps), and how to
+configure and run them as an ADM.
+
+**NOTE** - This document assumes you're generally familiar with the
+   system, and how we're using Hydra
+
+## Anatomy of a Pipeline ADM
```
jadie1, line 14: The random ADM makes sense as a starting example but I am wondering if the documentation needs a slightly more complex example as well to help demonstrate how to specify aspects like the attribute, structured inference engine, scenario description, prompt, and ICL. I think this readme is pretty clear but if I were referencing it to understand something like the `pipeline_comparative_regression.yaml` I might have remaining questions

jadie1 (approved):
Overall I think this looks really good and I am not opposed to merging as is. I think the documentation could be expanded slightly but we could work on that down the road

PR #166, opened 2025-03-14
Update package versions and integration test expected outputs

dmjoy:
This updates us to the latest available package versions for `transformers` (`4.49.0`) and `outlines` (`0.2.1`).  Note that the outputs are NOT identical compared to what our system generated before (see the changes to the integration test expected output files).  If we're to merge this we should not expect our Phase 1 Evalution ADMs to produce the same output (and scores) as before.  FWIW, you can always check out the previous version of the code if generating the same outputs is critical.  Also we'll need to clear out the outlines cache most likely as I think the format of that changed between versions.

I do think we'll have to accept this kind of drift to keep up with the latest available LLMs (via `transformers`), but would like to hear if there are major objections from folks on merging.

### DIFF
--- a/align_system/prompt_engineering/outlines_prompts.py
+++ b/align_system/prompt_engineering/outlines_prompts.py
@@ -174,6 +174,7 @@ def scenario_state_description_2(scenario_state):
     {% for event in scenario_state.events %}
     - {{ event.unstructured.rstrip() }}
     {% endfor %}
+    {% endif %}
     """
 
 @outlines.prompt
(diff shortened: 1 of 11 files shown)


dmjoy:
FWIW, my best guess as to why the results are different is there's an additional call to `random` somewhere in the updated transformers/outlines code that's meaning we don't get the same series of random outputs.  For the comparative regression outputs it looks like the estimated KDMA values are the same, but the random choice between them changed.  For the outlines structured baseline, it seems like the ordering of the actions changed (seems like they're ordered by the `unstructured` field, but for several actions that field is blank, so the ordering would be random between them).

PR #168, opened 2025-03-17
Add NAACL24 dataset files with adm, alignment, experiment configs

barry-ravichandran:
FYSA, @dmjoy @eveenhuis @jadie1 

This PR adds in the NAACL 24 formatted dataset files, configs and the code changes required for the outlines structured ADM to support the dataset attributes. PTAL when you get a chance and let me know your thoughts/feedback - thanks!


align_system/prompt_engineering/kdma_descriptions.yml
```diff
@@ -39,3 +39,40 @@ QualityOfLife:
   valid_scores:
     values: [1,3,7,9]
   relevant_structured_character_info: [demographics.age, injuries, vitals]
+
+# NAACL24 paper dataset
```
dmjoy, line 43: If possible I would like to try splitting these out into a separate file (maybe something like `naacl_kdma_descriptions.yml`), then they could be specified in the `inference_kwargs` for the config (I'll point it out in another comment).  Only because I don't want these entries to somehow interfere with our existing ADM's that use this file (not certain it would, but I think it makes sense to separate these).

align_system/configs/experiment/demo/outlines_structured_adm_aligned_naacl24.yaml
```diff
@@ -0,0 +1,26 @@
...
+adm:
+  instance:
+    precision: half
+    model_name: mistralai/Mistral-7B-Instruct-v0.3
+
+force_determinism: true
+align_to_target: true
+save_last_unstructured_state_per_scenario: true
+
+inference_kwargs:
```
dmjoy, line 23: In here you could set `kdma_descriptions_map: 'align_system/prompt_engineering/kdma_descriptions.yml'`

align_system/algorithms/outlines_adm.py
```diff
@@ -121,6 +131,36 @@ def kdma_value_to_system_prompt(kdma, value):
                 return low_maximization_system_prompt()
             else:
                 return high_maximization_system_prompt()
+        elif kdma == "ProtocolFocus":
+            if value == 0.0:
+                return low_protocol_focus_system_prompt()
+            else:
+                return high_protocol_focus_system_prompt()
```
dmjoy, line 138: Curious why we're only using the "low" prompt when the value is `0.0` (instead of splitting the range in half, e.g. `< 0.5` as above)?

dmjoy (commented):
Overall I think these changes are fine (happy to see there's minimal code changes needed); a few questions and possible tweaks to make.  Let me know what you think

align_system/algorithms/outlines_adm.py
barry-ravichandran: The only values for the NAACL24 KDMAs are 0.0 and 10.0 for high and low respectively. But the `fairness` attribute has a value of 5.0 to indicate a balanced decision. But at the moment, we are only focusing on the high and low prompt cases here.

align_system/algorithms/outlines_adm.py
```diff
@@ -86,6 +86,10 @@ def __init__(self,
 
             model_kwargs['torch_dtype'] = torch_dtype
 
+        if 'trust_remote_code' in kwargs:
+            if kwargs['trust_remote_code'] is True:
+                model_kwargs['trust_remote_code'] = True
+
```
dmjoy, line 92: I would prefer not to "officially" support this parameter.  I will follow up with how I set these from the config

align_system/algorithms/outlines_adm.py
dmjoy: Looking to future proof this a little bit (I would set it to a `< 0.5` as low, and otherwise set it to high) the current set up is not intuitive and I can see it biting us in the future when we're wondering why we're getting a high prompt when we have it set to `0.1` if we ever re-run experiments around this stuff

align_system/configs/experiment/demo/outlines_structured_adm_aligned_naacl24.yaml
```diff
@@ -0,0 +1,31 @@
...
+  # 'example_data/input_output_files/NAACL24_dataset_split/oracle_risk_aversion_training_input_output.json'
+  # 'example_data/input_output_files/NAACL24_dataset_split/oracle_utilitarianism_training_input_output.json'
+adm:
+  instance:
+    precision: half
+    model_name: mistralai/Mistral-7B-Instruct-v0.3
+    model_kwargs:
+      trust_remote_code: true
+    tokenizer_kwargs:
+      trust_remote_code: true
```
dmjoy, line 21: Sorry last nit-pick here; Let's comment these out as the default (could have a comment above mentioning when it might be necessary)

align_system/configs/experiment/demo/outlines_structured_adm_baseline_naacl24.yaml
```diff
@@ -0,0 +1,25 @@
...
+  # 'example_data/input_output_files/NAACL24_dataset_split/oracle_utilitarianism_training_input_output.json'
+adm:
+  instance:
+    precision: half
+    model_name: mistralai/Mistral-7B-Instruct-v0.3
+    baseline: true
+    model_kwargs:
+      trust_remote_code: true
+    tokenizer_kwargs:
+      trust_remote_code: true
```
dmjoy, line 22: Same request here to comment these out by default.

dmjoy (changes requested):
LGTM, please address the two comments though (commenting out the `trust_remote_code` stuff by default) then I'll be happy to merge.

PR #170, opened 2025-03-20
Add config for naacl24 kaleido runs

dmjoy:
@barry-ravichandran here's the few tweaks needed to get Kaleido up and running on the NAACL24 data.

Note that I'm using short descriptions of the KDMAs (rather than the longer descriptions specified in the other naacl24 KDMA descriptions file you added) I believe way back at the time of the NAACL24 paper I had tried both long and short versions, and these shorter versions worked a little better.


align_system/algorithms/lib/templates/kdma_descriptions_short_naacl24_paper.yml
dmjoy, line 1: Note that this file is specific to Kaleido (and isn't the same as the KDMA descriptions we normally use; will be something for me to figure out how to consolidate long term)

barry-ravichandran (approved):
LGTM!

PR #171, opened 2025-03-25
Fix scenario state description across all NAACL24 dataset KDMAs

barry-ravichandran:
@dmjoy @PaulHax This MR pushes in the fix for the NAACL24 ALIGN API dataset format.

In addition, this MR also addresses removing a scenario record from the `fairness` dataset which does not have the `fairness` attribute association linked to it.

Thanks @PaulHax for finding this bug in the dataset.

@brianhhu @jadie1 @eveenhuis Looping you in for situational awareness.


example_data/input_output_files/NAACL24_dataset_split/oracle_fairness_training_input_output.json
```diff
@@ -590,188 +590,11 @@
             }
         ]
     },
-    {
```
dmjoy, line 593: Not sure if this is a mirage or what, but curious why this probe seems to have been removed?

dmjoy: My mistake, didn't see the initial comment clarifying why it was removed.  LGTM

dmjoy (approved):

PR #172, opened 2025-03-25
 Factor get_dialogs to static method in outlines_adm

PaulHax:
Just leaving this here for now...

Mostly copy pastaed from #157 but made `get_dialogs` static so I don't have to load LLM to generate system prompt.

### DIFF
--- a/align_system/algorithms/outlines_adm.py
+++ b/align_system/algorithms/outlines_adm.py
@@ -1,10 +1,10 @@
 import json
 import random
 import itertools
-import numpy as np
 import torch
 import yaml
 import copy
+from functools import partial
 
 import outlines
 from outlines.samplers import MultinomialSampler
@@ -25,7 +25,6 @@ from align_system.utils.voting import (
     calculate_votes,
     filter_votes_to_responses,
 )
-from align_system.utils.hydrate_state import hydrate_scenario_state
 from align_system.algorithms.abstracts import ActionBasedADM
 from align_system.prompt_engineering.outlines_prompts import (
     baseline_system_prompt,
@@ -46,8 +45,6 @@ from align_system.prompt_engineering.outlines_prompts import (
     tag_choice_json_schema,
     treatment_choice_json_schema,
     treatment_choice_from_list_json_schema,
-    detailed_unstructured_treatment_action_text,
-    detailed_unstructured_tagging_action_text,
     high_risk_aversion_system_prompt,
     low_risk_aversion_system_prompt,
     high_continuing_care_system_prompt,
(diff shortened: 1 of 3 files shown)


align_system/algorithms/outlines_adm.py
```diff
@@ -349,24 +361,73 @@ def top_level_choose_action(self,
...
         # machine from looping indefinitely in some cases, see:
         # https://github.com/outlines-dev/outlines/issues/690#issuecomment-2102291934
         generator = outlines.generate.json(
             self.model,
-            action_choice_json_schema(json.dumps(choices)),
+            action_choice_json_schema(json.dumps(choices), reasoning_max_length),
             sampler=self.sampler,
             whitespace_pattern=r"[ ]?")
 
+        if max_generator_tokens >= 0:
```
dmjoy, line 422: This is an interesting approach, normally I would set the default in the kwargs to `None` and then if it's not `None` pass the value along as-is; but I guess in this case (setting the default to `-1`) you get to do some semantic validation on the value at the same time.  Is that kind of the intention or is this just a tomato vs tomato.

PaulHax: The -1 is a gross hack, and I regret it already =)

align_system/algorithms/outlines_adm.py
```diff
@@ -279,8 +292,8 @@ def top_level_choose_action(self,
                 kdma_descriptions = yaml.load(f, Loader=yaml.FullLoader)
             name = kdma_descriptions[kdma]['name']
 
-            positive_system_prompt = self.__class__.kdma_value_to_system_prompt(kdma, value)
-            negative_system_prompt = self.__class__.kdma_value_to_system_prompt(kdma, negative_value)
+            positive_system_prompt = OutlinesTransformersADM.kdma_value_to_system_prompt(kdma, value)
+            negative_system_prompt = OutlinesTransformersADM.kdma_value_to_system_prompt(kdma, negative_value)
```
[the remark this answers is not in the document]
dmjoy: Ah, I missed that bit (not sure about getting `__class__` in a static method) let's leave as is

align_system/algorithms/outlines_adm.py
dmjoy: I wouldn't say it's a hack, just curious about the tradeoffs; I'm content to leave it as is.

dmjoy (approved):
Overall LGTM, happy to merge as is if you're ready.

PaulHax:
OK, good to go by me, one less thing off my plate...

PR #173, opened 2025-04-01
Fix dataset validation errors

PaulHax:
By pinning commit SHA of itm-evaluation-client, aka swagger-client.

There is a bug due to upstream itm-evaluation-client changes.  Steps

```
python3 -m venv .venv
source .venv/bin/activate
pip install -e .
run_align_system
```

Get error
```
Error executing job with overrides: []
Error in call to target 'align_system.interfaces.input_output_file.InputOutputFileInterface':
TypeError('swagger_client.models.meta_info.MetaInfo() argument after ** must be a mapping, not MetaInfo')
full_key: interface
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ requests = "^2.31.0"
 bert-score = "^0.3.13"
 rich = "^13.6.0"
 rouge-score = "^0.1.2"
-swagger-client = {git = "https://github.com/NextCenturyCorporation/itm-evaluation-client.git", rev = "development"}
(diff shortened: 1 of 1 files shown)


barry-ravichandran (approved):

dmjoy:
Thanks Paul, yes I think this change makes sense in the near term.  Before approving / merging I just want to double check on my end that this is the right version to pin to

dmjoy:
Confirmed with the TA3 team that pinning to major versions of the client code probably makes the most sense.  Let's change the `rev` here to the tag `0.4.0` and I'll be happy to merge (that's the last version they officially released).

PaulHax:
Changed to `0.4.0`

dmjoy (approved):

PR #180, opened 2025-05-05
Update pipeline ADM documentation

dmjoy:
Update the Pipeline ADM docs to explain more complicated components etc. (using the comparative regression pipeline example).


dmjoy:
@jadie1 at your leisure could you go through the documentation additions here and let me know if you feel like this checks the box for you, or if there are more things you'd like to see added.

PR #183, opened 2025-05-20
Phase 2 Medical Urgency Alignment Function

eveenhuis:


align_system/algorithms/alignment_adm_component.py
```diff
@@ -47,3 +49,94 @@ def run(self,
...
+                "Assumption violated: `alignment_target` was None"
+            )
+
+        target_kdmas = alignment_target_to_attribute_targets(
+            alignment_target,
+            self.attributes)
+
+        target_kdmas = [dict(t) for t in target_kdmas]
+        if len(target_kdmas) > 1:
+            raise RuntimeError("Multi-kdma alignment has not yet been implemented")
```
dmjoy, line 85: Nit: there's a `NotImplementedError` exception type that seems slightly more appropriate here (and below)

align_system/algorithms/alignment_adm_component.py
```diff
@@ -47,3 +49,94 @@ def run(self,
...
+
+        def _handle_single_value(predictions):
+            if not isinstance(predictions, list):
+                return [predictions]
+            return predictions
+
+        choices = list(attribute_prediction_scores.keys())
+        if len(choices) == 1:
+            return choices[0]
```
dmjoy, line 94: Are we ever expecting only a single choice?  Might be good to log something here as a breadcrumb trail

align_system/algorithms/alignment_adm_component.py
```diff
@@ -47,3 +49,94 @@ def run(self,
...
+        # Equation from ADEPT
+        probe_midpoint = 0.5 + (medical_delta - attribute_delta)/2
+
+        attr_target = target_kdmas[attr]["value"]
+        if attr_target < probe_midpoint:
+            return opt_a["choice"]
+        elif attr_target > probe_midpoint:
+            return opt_b["choice"]
+        else:  # Midpoint == target, choose randomly
+            return random.choice([opt_a["choice"], opt_b["choice"]])
```
dmjoy, line 142: Again good to log something here imo

align_system/algorithms/alignment_adm_component.py
```diff
@@ -47,3 +49,94 @@ def run(self,
...
+
+        # Guaranteed to only have 1 attribute due to earlier checks
+        attr = target_kdmas[0]["kdma"]
+        attribute_delta = opt_a["kdmas"][attr] - opt_b["kdmas"][attr]
+        if attribute_delta > 0:
+            # opt_a is more medically urgent AND more deserving based on attribute
+            return opt_a["choice"]
+
+        # Equation from ADEPT
+        probe_midpoint = 0.5 + (medical_delta - attribute_delta)/2
```
dmjoy, line 134: I would want to know in the logs what the `probe_midpoint` is (maybe with `log.explain` with some text around it; could even say something about how the target is above or below the midpoint)

dmjoy (commented):
This generally looks fine, I just made a few suggestions, mainly around logging.  This might also be a good function to have a unit test (or few) for since it's a pretty critical component.  Hoping this is easier to do now with the pipeline components.

dmjoy (approved):

PR #184, opened 2025-05-22
Add relevance to midpoint alignment function

eveenhuis:
@jadie1 @dmjoy @barry-ravichandran 

Adds probe-level relevance (weighting) to midpoint alignment function and extends it to multi-kdma.

If provided, relevance values (should be 0-1) are used as follows in the midpoint equation

```
midpoint = 0.5 + (medical_weight * medical_delta - attr1_weight * attr1_delta ... - attrN_weight * attrN_delta) / total_weight
```

when relevance is not provided this simplifies to (+1 in denominator is for medical weight)

```
midpoint = 0.5 + (medical_delta - attr1_delta ... - attrN_delta)/(N_targets + 1)
```

and when only 1 target KDMA is provided this further simplifies to the original equation

```
midpoint = 0.5 + (medical_delta - attr1_delta)/2
```

If only binary relevance values are provided, the equations should still work, a relevance of 0 will act like the KDMA is not considered at all :)

*Note that I factored out the negative when summing across deltas, so that's why some of the comparisons (and the equation) flipped sign*


dmjoy:
I think we can address the non-binary relevance question (and the best way to handle that) after the 28th delivery.  I'm not sure that using relevance to weight the deltas is the right way to go; I was thinking using the relevance as more of a weight on the distance between the decision point and the alignment target.  Thinking of it like the distance between the target and the decision point as the "strength" of the pull in one direction or the other, if the attributes are both pulling in the same direction (e.g. to the same choice) then we just go with that choice.  If they're pulling in opposite directions, we would weight by relevance to determine the "stronger" pull.  (That said, I suppose there are multiple ways to tackle this problem, but I think we can largely punt on it for right now).

Also I think this is the kind of thing we really need tests for, because looking at the code and your explanation I believe it's correct and true to how it's been described, but the way to really verify that (repeatedly) is with unit tests.  The tests can also communicate exactly what happens in the corner cases (and forces us to define the behavior in those cases as well).

eveenhuis:
@dmjoy I went with this approach as it was similar to what ADEPT was doing with weighting in their server code, but we can certainly try out multiple approaches and see which one preforms the best

Unit testing added as requested

As always, let me know if there are questions/concerns. Thanks!

PR #191, opened 2025-05-26
Ph2 ADM tweaks for Kaleido "mashup"

dmjoy:
WIP

### DIFF
--- a/align_system/algorithms/icl_adm_component.py
+++ b/align_system/algorithms/icl_adm_component.py
@@ -38,7 +38,7 @@ class ICLADMComponent(ADMComponent):
                  scenario_description_template,
                  prompt_template,
                  attributes=None,
-                 predict_medical_urgency=False):
+                 target_attribute_names_override=None):
         self.icl_generator_partial = icl_generator_partial
         self.scenario_description_template = scenario_description_template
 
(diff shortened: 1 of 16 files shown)


align_system/algorithms/icl_adm_component.py
```diff
@@ -72,22 +81,26 @@ def run(self,
...
-                    # tuple
-                    kdma_values = ((attribute.kdma, target_kdma_value['value']),)
-                else:
-                    if self.predict_medical_urgency:
-                        kdma_values = (('medical', 1.0),)
+
+            # Not sure how much this value actually matters for ICL;
+            # defaulting to `1.0` if not in the alignment target
+            # (e.g. for the "medical" attribute)
+            value_for_attribute = alignment_target_value_lookup.get(attribute.kdma, 1.0)
```
eveenhuis, line 95: I don't think it's needed for this PR but if we realize it's important, we could accept a dictionary instead of a list for this override parameter

eveenhuis (approved):
I don't think I'm going to be able to provide a better review than Jadie actually running and verifying things :) LGTM!

align_system/algorithms/icl_adm_component.py
dmjoy: Yes I had the same series of thoughts

PR #192, opened 2025-05-26
Delta based relevance for comp reg alignment to multi-attribute targets

jadie1:
Enables comp. reg. alignment to multi-attribute targets by adding a component step that for each probe:
1. Finds the attribute with smaller predicted delta and removes it from predicted scores
2. Alters the alignment target to only include the more relevant (bigger delta) attribute

Example call to run zeroshot:
```
run_align_system +experiment=phase2_june_collab/multi_attribute_pipeline_zeroshot_comparative_regression +alignment_target=june2025/ADEPT-June2025-affiliation_merit-0.0_0.0.yaml
```

Example call to run fewshot:
```
run_align_system +experiment=phase2_june_collab/multi_attribute_pipeline_fewshot_comparative_regression_loo +alignment_target=june2025/ADEPT-June2025-affiliation_merit-0.0_0.0.yaml
```

### DIFF
--- a/align_system/algorithms/misc_itm_adm_components.py
+++ b/align_system/algorithms/misc_itm_adm_components.py
@@ -172,11 +172,10 @@ class Phase2RegressionRemoveIrrelevantAttributes(ADMComponent):
 
     def run(self,
             attribute_prediction_scores,
-            alignment_target
-            ):
+            alignment_target):
         # If there are two non-medical attributes, removes the one with smaller delta
 
-        attributes = list(next(iter(attribute_prediction_scores.values())).keys())
+        attributes = list({key for inner in attribute_prediction_scores.values() for key in inner})
         # Ignore / don't filter out medical
         keep_attributes = []
         if 'medical' in attributes:


align_system/algorithms/misc_itm_adm_components.py
```diff
@@ -162,3 +163,55 @@ def run_returns(self):
...
+
+class Phase2RegressionRemoveIrrelevantAttributes(ADMComponent):
+    def run_returns(self):
+        return ('attribute_prediction_scores',
+                'alignment_target')
+
+    def run(self,
+            attribute_prediction_scores,
+            alignment_target
+            ):
```
dmjoy, line 176: Slight preference for putting this `)` at the end of the above line.

align_system/algorithms/misc_itm_adm_components.py
```diff
@@ -162,3 +163,55 @@ def run_returns(self):
...
+        # If there are two non-medical attributes, removes the one with smaller delta
+
+        attributes = list(next(iter(attribute_prediction_scores.values())).keys())
```
dmjoy, line 179: This probably works fine most of the time, the safest thing would be to build up a `set` of all keys for the values so we don't miss any.  (That said maybe it's an error if we don't have the same attributes for all choices)

dmjoy (approved):

align_system/algorithms/misc_itm_adm_components.py
```diff
@@ -162,3 +163,55 @@ def run_returns(self):
...
+        # Ignore / don't filter out medical
+        if 'medical' in attributes:
+            attributes.remove('medical')
+
+        # Only one attribute aside from medical -> filtering not needed
+        if len(attributes) == 1:
+            return attribute_prediction_scores, alignment_target
+
+        # Two attributes -> remove the one with smaller delta
+        elif len(attributes) == 2:
```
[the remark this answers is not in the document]
jadie1: Agreed, I've fixed this

align_system/algorithms/misc_itm_adm_components.py
```diff
@@ -162,3 +163,57 @@ def run_returns(self):
...
+            keep_attributes.append('medical')
+
+        # Only one attribute aside from medical -> filtering not needed
+        if len(attributes) == 1:
+            return attribute_prediction_scores, alignment_target
+
+        # Two or more attributes -> keep the one with largest delta
+        else:
+            if len(attribute_prediction_scores.keys()) > 2:
+                raise RuntimeError("Relevance filtering not implemented for more than two choices.")
```
eveenhuis, line 192: Nit: David recommended I raise `NotImplementedError` for a similar check

align_system/algorithms/misc_itm_adm_components.py
```diff
@@ -162,3 +163,57 @@ def run_returns(self):
...
+    def run_returns(self):
+        return ('attribute_prediction_scores',
+                'alignment_target')
+
+    def run(self,
+            attribute_prediction_scores,
+            alignment_target):
+        # If there are two non-medical attributes, removes the one with smaller delta
+
+        attributes = list({key for inner in attribute_prediction_scores.values() for key in inner})
```
eveenhuis, line 178: I assume we're looping all choice predictions in case different choices have different predictions? Do we think that'll actually happen?

align_system/algorithms/misc_itm_adm_components.py
```diff
@@ -162,3 +163,57 @@ def run_returns(self):
...
+        # Two or more attributes -> keep the one with largest delta
+        else:
+            if len(attribute_prediction_scores.keys()) > 2:
+                raise RuntimeError("Relevance filtering not implemented for more than two choices.")
+            else:
+                # Determine most relevant attribute to keep
+                choiceA, choiceB = list(attribute_prediction_scores.keys())
+                max_delta = -np.inf
+                for attr in attributes:
+                    delta = abs(np.array(attribute_prediction_scores[choiceA][attr]).mean() - np.array(attribute_prediction_scores[choiceB][attr]).mean())
```
eveenhuis, line 198: I know at one point we potentially had to handle both a single prediction and a list of predictions. Do we still need to do that? (If not yayyyyy simpler code :))

align_system/algorithms/misc_itm_adm_components.py
```diff
@@ -162,3 +163,57 @@ def run_returns(self):
...
+                    if delta > max_delta:
```
eveenhuis, line 199: This will only keep the first one if there's a tie. I think that's fine because I don't know what we'd actually do if there was a tie (plus that seems unlikely), but just double checking you didn't have a different set of assumptions

align_system/algorithms/misc_itm_adm_components.py
```diff
@@ -162,3 +163,57 @@ def run_returns(self):
...
+                        max_delta = delta
+                        relevant_attribute = attr
+                keep_attributes.append(relevant_attribute)
+
+                # Update predicted scores to only have more relevant attribute
+                filtered_attribute_prediction_scores = {choiceA:{}, choiceB:{}}
+                for keep_attr in keep_attributes:
+                    filtered_attribute_prediction_scores[choiceA][keep_attr] = attribute_prediction_scores[choiceA][keep_attr]
+                    filtered_attribute_prediction_scores[choiceB][keep_attr] = attribute_prediction_scores[choiceB][keep_attr]
```
eveenhuis, line 208: (nit)

```suggestion
                # Update predicted scores to only have more relevant attribute
                filtered_attribute_prediction_scores = {
	            choice: {
    	                keep_attr: predictions[choice][keep_attr]
    	                for keep_attr in keep_attributes
                    }
	            for choice in attribute_prediction_scores.keys()
                }
```

eveenhuis (approved):
Overall LGTM, I'd be fine with merging as is. Just a couple of nits/sanity checks

PR #194, opened 2025-05-27
Relevance Oracle Pipeline

eveenhuis:
Adds oracle relevance pipeline and component

### DIFF
--- a/align_system/configs/adm/phase2_pipeline_relevance_oracle.yaml
+++ b/align_system/configs/adm/phase2_pipeline_relevance_oracle.yaml
@@ -1,0 +1,40 @@
+name: phase2_pipeline_relevance_oracle
+
+defaults:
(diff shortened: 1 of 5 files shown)


align_system/configs/adm/phase2_pipeline_relevance_oracle.yaml
```diff
@@ -0,0 +1,40 @@
+name: phase2_pipeline_regression_oracle
```
jadie1, line 1: Nit: may want to change name phase2_pipeline_relevance_oracle

jadie1 (approved):

align_system/configs/adm/phase2_pipeline_relevance_oracle.yaml
eveenhuis: Agh I did this so many times accidentally, must have missed this one

PR #206, opened 2025-06-04
Add caching to comparative regression pipeline component

dmjoy:
To turn on caching for the comparative regression component, in the `step_definition` for `comparative_regression` set `enable_caching: true` (it's off / `false` by default) e.g.:
```YAML
step_definitions:
  comparative_regression:
    scenario_description_template: ${ref:adm.scenario_description_template}
    prompt_template: ${ref:adm.prompt_template}
    score_schema_template: ${adm.comparative_regression_choice_schema}

    enable_caching: true
```

This uses the `ubelt` package (new dependency) and caches to disk (into this directory: `~/.cache/ubelt/`, in case you ever need to clear it).

I believe I've captured all of the "dependencies" for comparative regression for caching, but please let me know if I've missed anything.


align_system/algorithms/comparative_regression_adm_component.py
```diff
@@ -136,10 +168,47 @@ def run(self,
                         attribute_prediction_reasonings[choice].setdefault(
                             attribute.kdma, []).append(response[choice]['reasoning'])
                     # Probe level reasoning
-                    except:
+                    except KeyError:
```
dmjoy, line 171: I'm assuming this is the error we want to `except` here, but someone correct me if I'm wrong (always better to be as specific as possible when using `except`)

PR #208, opened 2025-06-07
Prepend the scenario unstructured to the scene/probe unstructured

dmjoy:
Phase2 data has been updated to move the scenario description out from the `state.unstructured` field into the unstructured field for the scenario (and it's been put into the `threat_state.unstructured` as a convenience for us).  This PR "hacks" the unstructured state text coming back from the server to put that scenario text in front again (so the Phase2 ADMs don't need to change anything).


dmjoy:
FYSA @eveenhuis @jadie1

PR #212, opened 2025-07-15
Update medical urgency alignment function

eveenhuis:
Updates medical urgency function based on ADEPT's changes.

Moves relevance consideration to voting weights instead of actually affecting midpoint calculation.

### DIFF
--- a/align_system/algorithms/alignment_adm_component.py
+++ b/align_system/algorithms/alignment_adm_component.py
@@ -126,6 +126,169 @@ class MedicalUrgencyAlignmentADMComponent(ADMComponent):
     def run_returns(self):
         return ('chosen_choice', 'best_sample_idx')
 
+    def _midpoint_eqn(self, medical_delta, attribute_delta, total_weight=2):
+        # Midpoint equation from ADEPT
+        return 0.5 + (medical_delta - attribute_delta)/total_weight
+
+    def run(
+        self,
+        attribute_prediction_scores,
+        alignment_target,
+        attribute_relevance=None,
+    ):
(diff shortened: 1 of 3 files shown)


align_system/algorithms/alignment_adm_component.py
```diff
@@ -126,9 +126,17 @@ def __init__(
...
+        medical_weight = medical_weights.get(kdma, 1.0)
+
         # Midpoint equation from ADEPT
-        return 0.5 + (medical_delta - attribute_delta)/total_weight
+        if kdma == "affiliation":
+            return (opt_a_value + medical_weight*medical_delta - opt_a_value*medical_delta)/2
+        elif kdma == "merit":
+            return (opt_a_value + medical_weight*medical_delta)/4
+        else:
+            return 0.5 + (medical_weight*medical_delta - attr_delta)/2
```
dmjoy, line 139: I would much rather these weights be parameterized, but since there are different formulas for some of the attributes anyway I suppose it's more understandable to these hardcoded bits together in one place.  (No changes requested here)

dmjoy (changes requested):
I think these changes are largely fine and in line with what Dustin posted.  One change I would like to see here is keeping the old midpoint alignment function implementation as it was, and adding this new one under a new class, maybe `MedicalUrgencyAlignmentWeightedADMComponent` (or any equally long name) or `MedicalUrgencyAlignmentV2ADMComponent`.  Just so we can keep backward compatibility (since I don't think we can replicate the old alignment behavior with this new implementation)

PR #215, opened 2025-07-15
Reduce code duplication between medical urgency alignment functions

eveenhuis:
Potential refactor for the medical urgency alignment functions. Not 100% consistent with previous behavior but the pieces that changed were unused. Those pieces are:

1. Relevance now affects voting weight instead of affecting the midpoint directly (I think this is what @dmjoy  might have been arguing for when the initial alignment function was merged). Regardless, we've only used binary relevance so far, so we weren't really hitting this functionality. [unit tests still pass, with the exception of the 2 non-binary relevance tests which are now updated for this new behavior]
2. Removal of on-the-fly medical urgency relevance prediction (`medical_weight`). The previous function now just uses a value of 1.0, the weighted class uses the weights specified by ADEPT. Based on ADEPT's updates, changing the medical_weight does things to the midpoint computation that we were not accounting for.  [unit tests removed]

I also parameterized the test class so that all of `MedicalUrgencyAlignmentADMComponent`'s tests are also run with `MedicalUrgencyAlignmentWeightedADMComponent` so that we don't have to duplicate those test cases. We  then also have the specialized tests for the weighted midpoint equations for affiliation and merit.

I understand if we don't want to merge this in, but thought I'd offer it.

### DIFF
--- a/align_system/algorithms/alignment_adm_component.py
+++ b/align_system/algorithms/alignment_adm_component.py
@@ -126,180 +126,9 @@ class MedicalUrgencyAlignmentADMComponent(ADMComponent):
     def run_returns(self):
         return ('chosen_choice', 'best_sample_idx')
 
-    def _midpoint_eqn(self, medical_delta, attribute_delta, total_weight=2):
-        # Midpoint equation from ADEPT
-        return 0.5 + (medical_delta - attribute_delta)/total_weight
-
-    def run(
-        self,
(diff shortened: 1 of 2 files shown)


dmjoy:
Coming back to this, I'm running a quick test on this branch to make sure we get the same scores as before (using the training data) for the comp reg ADM.  Assuming all is well there I'll merge

PR #217, opened 2025-07-24
Remove non-determinism from midpoint alignment functions

eveenhuis:
- If target is exactly the midpoint, we now choose the medically-needy patient, instead of not voting for that KDMA (which made it a random choice in the single KDMA case)
- Additionally, if multiple patients receive the same number of votes, we now choose the "first" patient instead of randomly choosing, to remove that source of non-determinism
- There is still one case where we don't vote: both patients receive the same medical *and* same attribute scores. In the multi-KDMA case, this is equivalent to ignoring that KDMA (decision would be based on the other KDMA votes). In the single KDMA case, neither patient would receive a vote, i.e. tied at 0 votes, so we fall back to choosing the first patient as described in the previous bullet

### DIFF
--- a/align_system/algorithms/alignment_adm_component.py
+++ b/align_system/algorithms/alignment_adm_component.py
@@ -1,5 +1,4 @@
 import math
-import random
 
 from align_system.utils import call_with_coerced_args, logging
 from align_system.algorithms.abstracts import ADMComponent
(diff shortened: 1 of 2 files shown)


dmjoy (approved):
LGTM thanks

PR #220, opened 2025-07-25
Instructions for hugging face token cache

DennisMelamedKitware:
Added instructions to README to set up a hugging face personal access token, allows accessing 'gated' models for which you have to accept the T&Cs on the hugging face website.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,31 @@ RAM and with a modern GPU with at least 12GB of memory.
 It's generally recommended to set up a virtual Python environment to neatly manage dependencies (e.g. using `venv` or `conda`).  The `align-system` code can be installed as a Python module with `pip
 install git+https://github.com/ITM-Kitware/align-system.git`.
 
+Some huggingface models are 'gated' and require accepting terms and conditions (e.g. [Mistral-7B-Instruct-v0.2](https://huggingface.co/mistralai/Mistral-7B-Instruct-v0.2)). 
+A gated model is indicated by errors like this:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -12,6 +12,31 @@ RAM and with a modern GPU with at least 12GB of memory.
 It's generally recommended to set up a virtual Python environment to neatly manage dependencies (e.g. using `venv` or `conda`).  The `align-system` code can be installed as a Python module with `pip
 install git+https://github.com/ITM-Kitware/align-system.git`.
 
+Some huggingface models are 'gated' and require accepting terms and conditions (e.g. [Mistral-7B-Instruct-v0.2](https://huggingface.co/mistralai/Mistral-7B-Instruct-v0.2)). 
+A gated model is indicated by errors like this:
```
dmjoy, line 16: I think I would prefer moving this down just a bit to the "Running the system" section, with a lead-in like "Note that some huggingface models ..." which then points to a new section in the top-level README for "HuggingFace Token Setup" that contains the remaining detailed instructions.

(Note to self it's probably a better idea to make the default configuration for `run_align_system` not use a gated model, just to simplify the initial setup and smoketest)

dmjoy (changes requested):

README.md
DennisMelamedKitware: moved things around - hopefully this matches what you were aiming for.

dmjoy (approved):

PR #223, opened 2025-08-01
DecisionFlow integration

barry-ravichandran:
This MR handles the integration of the DecisionFlow algorithm based off of this source code: [DecisionFlow](https://github.com/xiusic/DecisionFlow)

### DIFF
--- a/align_system/algorithms/outlines_inference_engine.py
+++ b/align_system/algorithms/outlines_inference_engine.py
@@ -85,7 +85,7 @@ class OutlinesTransformersInferenceEngine(StructuredInferenceEngine):
             yield batch
 
     @classmethod
-    def run_in_batches(cls, inference_function, inputs, batch_size, max_generator_tokens):
+    def run_in_batches(cls, inference_function, inputs, batch_size, max_generator_tokens=None):
         ''' Batch inference to avoid out of memory error'''
         outputs = []
         for batch in cls.batched(inputs, batch_size):


align_system/configs/adm/outlines_decision_flow_aligned/pipeline_decision_flow.yaml
```diff
@@ -0,0 +1,37 @@
+name: pipeline_decision_flow
+
+defaults:
+  # Import defaults into this namspace (adm) as @name, for further
+  # customization
+
```
barry-ravichandran, line 6: add adm override here.

align_system/algorithms/decision_flow_adm/variable_stage_component.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def run(self, scenario_state, choices, **kwargs):
+        """Extract variables from task and choices"""
+
+        dialog = [
+            {"role": "system", "content": self.system_prompt},
+            {"role": "user", "content": self.prompt_template}
+        ]
+        response = self.structured_inference_engine.run_inference_unstructured(
+            self.structured_inference_engine.dialog_to_prompt(dialog),
```
barry-ravichandran, line 31: Revert back to using the way to get a valid json. (run_inference)

align_system/algorithms/decision_flow_adm/math_reason_stage_component.py
```diff
@@ -0,0 +1,179 @@
...
+            # Validate answer index
+            if not isinstance(answer_idx, int) or answer_idx < 0 or answer_idx >= len(actions):
+                log.warning(f"Invalid answer index {answer_idx}, defaulting to 0")
+                answer_idx = 0
+
+            chosen_action = actions[answer_idx]
+
+            # Add reasoning as justification if possible
+            if hasattr(chosen_action, 'justification') and chosen_action.justification is None:
+                chosen_action.justification = reasoning
```
dmjoy: This logic looks like it's from our pre-outlines ADM code.  There's probably a way to set up the schema such that we just have it pick the answer by it's text rather than the answer index.  But that's probably different from how the original decision flow is set up, so let's consider that for a follow-up MR.

PR #225, opened 2025-08-08
Make comparative regression reasoning length configurable

PaulHax:
Makes reasoning length configurable in comparative regression to allow downstream apps to override the hardcoded 512 character limit.  We want to avoid truncating the justification.

Alternativly...

Could remove the character length limit entirely and address out of memory errors the limit is ment to avoid by reducing batching of ICL samples.

### DIFF
--- a/align_system/prompt_engineering/outlines_prompts.py
+++ b/align_system/prompt_engineering/outlines_prompts.py
@@ -1316,14 +1316,14 @@ class ComparativeRegressionPrompt():
             attribute)
 
 
-def comparative_regression_json_schema(choices, scale_factor=100):
+def comparative_regression_json_schema(choices, scale_factor=100, reasoning_max_length=512):
     json_schema = {
         "type": "object",
         "properties": {
             "reasoning": {
                 "type": "string",
                 "minLength": 1,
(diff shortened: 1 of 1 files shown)


dmjoy:
I'm fine with these change as is, I did want to note the different options for getting longer (potentially untruncated) reasonings, though each of them have a downside:

1. As you've done here, you can increase the `maxLength` in the JSON schema; however when outlines builds out the FSM for the schema, the larger the `maxLength` the larger the final FSM will be and it takes longer and longer to compile it, which could be a noticeable slowdown
2. You can leave `maxLength` out of the schema entirely, which should let the LLM finish it's reasoning in every case; however we've seen seemingly infinite generations from the models in this case causing the code to become unresponsive
3. In the call to outlines (`outlines.generate.json()`) you can pass in a `max_tokens` argument which will cut off generation when that's hit; however there's no guarantee of a valid JSON output if you hit that `max_tokens` limit

All that said, if we leave the length at 512, the reasoning will get truncated in some cases.  My thought here is let's expose all of the options and the user can decide which way they would like the system to fail (in the rare event that it does).

PR #236, opened 2025-08-20
Experimental changes to allow choice swapping in the prompt

dmjoy:
**DO NOT MERGE IN CURRENT STATE**

A bit of hackery here to be able to reverse choices in the prompt for comparative regression.  There's also functionality here to be able to run two passes of comparative regression, one reversed and one not-reversed and concatenate them together (gets averaged by the alignment function) but I'm not very confident in the implementation so don't necessarily want to merge as-is.

@jadie1 FYSA


dmjoy:
Closing as these changes were merged in with #240

PR #240, opened 2025-09-02
Dev/choice ordering refactored

dmjoy:
Supplants #239 and #236.  This PR adds options to swap the ordering of choices for both the comparative regression and ICL examples (randomizing also a choice for ICL).  To accommodate running both swapped and not swapped and combining those regressed values (to be averaged downstream), there's a new hook added for an `output_conflict_resolver` function for each step that can specify how outputs should be handled if they already exist in the pipeline's `working_output` (by default new values replace old).

### DIFF
--- a/align_system/configs/adm/phase2_pipeline_fewshot_comparative_regression_swap_average.yaml
+++ b/align_system/configs/adm/phase2_pipeline_fewshot_comparative_regression_swap_average.yaml
@@ -1,0 +1,70 @@
+name: phase2_pipeline_zeroshot_comparative_regression_swap_average
+
+defaults:
+  # Import defaults into this namspace (adm) as @name, for further
+  # customization
+
+  # Shared variables / components
+  - /attribute@mu: medical_urgency
+  - /attribute@af: affiliation_focus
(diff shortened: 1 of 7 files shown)


align_system/configs/adm/phase2_pipeline_fewshot_comparative_regression_swap_average.yaml
```diff
@@ -0,0 +1,70 @@
...
+    prompt_template: ${ref:adm.prompt_template}
+    score_schema_template: ${adm.comparative_regression_choice_schema}
+
+  comparative_regression_swap:
+    scenario_description_template: ${ref:adm.scenario_description_template}
+    prompt_template: ${ref:adm.prompt_template}
+    score_schema_template: ${adm.comparative_regression_choice_schema}
+    reverse_choice_ordering: true
+    output_conflict_resolver:
+      _target_: align_system.algorithms.comparative_regression_adm_component.RegressionOutputsConflictResolver
```
dmjoy, line 54: Note that this is where that conflict resolver is getting set (only being used for this "swap_average" config)

PR #241, opened 2025-09-04
Add phase2 ICL examples to repository

### LINKED ISSUE #229: Public ICL examples location for library consuming code
Consider putting ICL examples somewhere that downstream library users can use.  Comparative Regression ADM hydra configs point to `/data/shared` on ITM machine. Downstream align-system users of CR ADM must download ICL examples from somewhere and then override ADM hydra config to point to local location.

Motivated by https://github.com/ITM-Kitware/align-app

PaulHax:
If we don't care to easly support running align-system off the ITM machine, we could not do this PR.  align-app is using https://github.com/PaulHax/align-data to get the ICL examples for the comparative regression ADM.

### Changes
- Copied a selection of medical domain training  JSONs from /data/shared/samba/
- Updated the .yaml in adm_component/icl to use local in repo paths.
- remove  `${oc.env:ICL_DATABASE_DIR,<default>}` environment variable pattern,  Prefer Hydra config overrides.

closes #229


dmjoy:
Thanks for putting this together @PaulHax , I don't think we need the `${oc.env:ICL_DATABASE_DIR,<default>}` anymore (for the reasons you mention).  I think I would prefer the more descriptive filenames here like: `June2025-MU-train_20250602.json` (just because even with the `phase2` subdirectory we might have more than one version for a given attribute that could be in play.

If you've got the will and the way, I do think having the other ICL files that are referenced committed here would be nice (just to pull the whole bandaid off at once)

Let's punt on a more sophisticated solution for now.

align_system/configs/adm_component/icl/phase2_comparative.yaml
```diff
@@ -16,10 +16,10 @@ icl_generator_partial:
...
-      medical: /data/shared/samba/phase2_icl/June2025-MU-train_20250602.json
-      affiliation: /data/shared/samba/phase2_icl/June2025-AF-train_20250523.json
-      merit: /data/shared/samba/phase2_icl/June2025-MF-train_20250523.json
-      personal_safety: /data/shared/samba/phase2_icl/June2025-PS-train_20250602.json
-      search: /data/shared/samba/phase2_icl/June2025-SS-train_20250527.json
+      medical: align_system/resources/icl/phase2/July2025-MU-train_20250804.json
+      affiliation: align_system/resources/icl/phase2/July2025-AF-train_20250804.json
+      merit: align_system/resources/icl/phase2/July2025-MF-train_20250804.json
+      personal_safety: align_system/resources/icl/phase2/July2025-PS-train_20250804.json
+      search: align_system/resources/icl/phase2/July2025-SS-train_20250804.json
```
PaulHax, line 23: These are udpated to the latest dated training json.  Not sure if that is a problem.

PaulHax: I added the June20205 JSON files `phase2_comparative.yaml` was pointing too.

PR #263, opened 2026-01-30
[DRAFT] Dev/OpenAI

ygefen:
Code & Config to utilize OpenAI API


align_system/algorithms/openai_inference_engine.py
```diff
@@ -0,0 +1,89 @@
+from typing import Union, Optional, Literal
+from functools import partial
+from copy import deepcopy
+from openai import OpenAI
+
+
+from align_system.algorithms.abstracts import StructuredInferenceEngine
+
+
+class VLLMInferenceEngine(StructuredInferenceEngine):
```
ygefen: Probably can delete, aside from the syntax error this introduces. Thought to maybe create two separate (perhaps parent/child) classes for OpenAI and VLLM to disambiguate the overlapping configuration parameters and to remove the need to check if api_key and/or base_url are set to determine implicitly if we are targeting actual OpenAI or and OpenAI compliant endpoint such as VLLM.

align_system/algorithms/openai_inference_engine.py
```diff
@@ -0,0 +1,89 @@
...
+    def dialog_to_prompt(self, dialog: list[dict]) -> str:
+        # OpenAI uses "developer" prompt instread of "system" (which is not exposed to the caller)
+        # https://platform.openai.com/docs/guides/prompt-engineering#message-roles-and-instruction-following
+        if not self.base_url:
+            # We are targetting the OpenAI service
+            prompt = deepcopy(dialog)
+            for p in prompt:
+                if p["role"] == "system":
+                    p["role"] = "developer"
+        return prompt
```
ygefen: I have a question about this.

So OpenAI API allows for a List[Dict] type prompt which is as returned here, but the signature in the base class and the example in [outlines_inference_engine.py](https://github.com/ITM-Kitware/align-system/blob/main/align_system/algorithms/outlines_inference_engine.py#L53) imply we want this flattened? 

We can do this in OpenAI by formatting as XML or Markdown as described here: https://platform.openai.com/docs/guides/prompt-engineering#message-formatting-with-markdown-and-xml 

Is this what you want? It would be more inline with the base class but would be bypassing a useful feature of OpenAI. I could always flatten the prompt returned from `dialog_to_prompt` and then rebuild the List[Dict] in `run_inference` if that is preferred.

align_system/configs/inference_engine/openai_gpt4.yaml
```diff
@@ -0,0 +1,18 @@
...
+inference_batch_size: 5
+
+# Optional kwargs
+base_url: null       # Set to null for OpenAI cloud
+api_key: null        # Reads from OPENAI_API_KEY env var if null
+organization: null   # Reads from OPENAI_ORG_ID  env var if null
+project: null        # Reads from OPENAI_PROJECT_ID  env var if null
+webhook_secret: null # Reads from OPENAI_WEBHOOK_SECRET  env var if null
+timeout: "NOT_GIVEN"
+# Optional unstable kwargs
```
ygefen: I added some extra fields not included in the example hydra config here: https://docs.google.com/document/d/1qrGrOhVj-0mbYJRU18G1fxEfFOBt0yJV0_oQTGklA6Q/edit?pli=1&tab=t.0

ygefen (commented):
@dmjoy Your thoughts would be appreciated

dmjoy (commented):
Looking good so far, and good questions.  Let me know if anything is still unclear.  I'm assuming that the extra requirements needed for the `openai` dependency are fairly light?  Assuming the answer is yes we can go with it for now, some part of me likes the idea of just using `requests` or some other generic HTTP request library to truly decouple, but that's probably not worth it.

dmjoy (commented):
In general this all looks fine though I did have a few questions and minor (I think) suggestions.  Also since this is an auxiliary component of the system and not necessarily / always critical path I'm not going to put too much scrutiny on it.

I would normally argue strongly for making the `uv` switch a separate PR, but since you need to add dependencies anyway, and I do also consider `uv` a strict upgrade to poetry I think it's fine to include here

align_system/algorithms/openai_inference_engine.py
```diff
@@ -0,0 +1,450 @@
...
+    first_content = message_output['content'][0]
+    text_content = first_content.get('text')
+
+    if text_content:
+        if has_schema:
+            # Parse as JSON for structured outputs
+            return json.loads(text_content)
+        else:
+            # Return as plain text for unstructured outputs
+            return {"response": text_content}
```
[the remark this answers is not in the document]
ygefen: actually for the structured output should i be returning `Dialog | DialogElement` objects instead of `Union[Dict, List[Dict]]?

dmjoy (changes requested):

align_system/prompt_engineering/outlines_prompts.py
```diff
@@ -1479,19 +1479,17 @@ def __call__(self, character, scenario_state):
...
+        if m:
             setup = m.group(1).strip()
         else:
-            # FIXME: This case is needed specifically for
-            # multi-attribute targets (and single attribute scenarios)
-            # where we may need to run regression for Personal Safety
-            # on an Affiliation scenario (where the text structure is
-            # a bit different)
-            return f"  - {char_unstructured}"
+            setup = collapsed
```
dmjoy: I would prefer to revert this back to the version I originally had in here.  I think the two approaches are almost identical except the original uses a non-greedy pattern for the "setup" string.

pyproject.toml
```diff
@@ -23,11 +23,27 @@ dependencies = [
...
-swagger-client = { git = "https://github.com/NextCenturyCorporation/itm-evaluation-client.git" }
+[tool.uv]
+find-links = ["https://download.pytorch.org/whl/cu118"]
+
+[tool.hatch.metadata]
+allow-direct-references = true
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
```
dmjoy, line 49: I'm not familiar with `hatchling`, what's the benefit / purpose of using it here? (No objection to including it, just curious)

dmjoy (changes requested):

dmjoy (approved):

PR #267, opened 2026-04-12
Minor UV changes

ygefen:
1. Creates a new dependency group for two dependencies that are neither explicit nor transitive dependencies, but are only used in a single file which is not imported elsewhere.
2. moves to the uv_build backend since we do not need the features of hatchling (this is not strictly necessary, but perhaps slightly cleaner)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,6 +34,10 @@ openai = [
 anthropic = [
     "anthropic>=0.45.0,<1.0.0",
 ]
+llama-index-retriever = [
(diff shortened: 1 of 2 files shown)


ygefen:
For a comparison on build backends: https://medium.com/@dynamicy/python-build-backends-in-2025-what-to-use-and-why-uv-build-vs-hatchling-vs-poetry-core-94dd6b92248f

dmjoy (approved):
LGTM

PR #268, opened 2026-04-13
Fix/dependabot

ygefen:


ygefen:
@dmjoy integ tests pass

dmjoy (approved):
LGTM and results on the training set are comparable.

PR #269, opened 2026-04-14
Add version tracking to input_output.json (closes #245)

ygefen:
Adds align_system/utils/version.py with get_version(), which uses setuptools-scm to produce an exact git-describe string at runtime (e.g. 0.5.10.dev3+g1234567), with a fallback to importlib.metadata for packaged/containerized environments.

The input_output.json format changes from a bare list to:
  {"version": "...", "results": [...]}

Integration test expected files are updated to match the new structure, and the "version" line is added to LOG_EXEMPTIONS in run_integration_test.py so the ever-changing version string is skipped during diff comparisons.

### DIFF
--- a/align_system/drivers/itm_phase1.py
+++ b/align_system/drivers/itm_phase1.py
@@ -9,6 +9,7 @@ from omegaconf import DictConfig, OmegaConf
 from timeit import default_timer as timer
 
 from align_system.utils import logging
+from align_system.utils.version import get_version
 from align_system.exceptions import SceneSkipException
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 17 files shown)


dmjoy:
LGTM.  I'd like to add one more piece of info to this `meta.json` file if possible.  The `username` defined here: https://github.com/ITM-Kitware/align-system/blob/be55f29e5832eba586df9eb59485bd9afc9c1aa1/align_system/interfaces/ta3_caci_action_based_service.py#L29

Thinking you should be able to access it as `interface.username`

dmjoy (approved):

PR #270, opened 2026-04-23
Adds RAG ADM Component and creates tagging pipeline to test it

ygefen:


ygefen:
[tagging_comparison_results.zip](https://github.com/user-attachments/files/27018799/tagging_comparison_results.zip)
Comparison between RAG pipeline and expert prompt pipeline for the tagging task

ygefen:
[tagging_comparison_results.zip](https://github.com/user-attachments/files/27129612/tagging_comparison_results.zip)
Results with the addition of `tagging fewshot aligned` experiments.

ygefen:
This code needs refactoring once I get a thumbs up on the concept and before merge. Specifically there is too much code repetition.

align_system/algorithms/outlines_baseline_adm_component.py
```diff
@@ -165,3 +165,100 @@ def _generic_object_repr(obj):
...
+                log.info("Cache hit for `outlines_rag_baseline_adm_component`"
+                         " returning cached output")
+                return cached_output
+            else:
+                log.info("Cache miss for `outlines_rag_baseline_adm_component` ..")
+
+        scenario_description = call_with_coerced_args(
+            self.scenario_description_template,
+            {'scenario_state': scenario_state,
+             'rag_context': rag_context})
```
dmjoy, line 207: Interesting, so is this really the only difference with the base class?  If so wondering if it just makes more sense to add it to the original class 🤔

dmjoy: Nvm.  Thinking it's better to be a distinct subclass but let's make `rag_context` a required argument to the `run` function (if you're not using RAG just use the base class?)

align_system/algorithms/prompt_based_aligned_adm_component.py
```diff
@@ -183,3 +183,144 @@ def run(self,
...
+    will silently ignore it via call_with_coerced_args.
+    """
+
+    def run(self,
+            scenario_state,
+            choices,
+            alignment_target,
+            positive_icl_dialog_elements=[],
+            negative_icl_dialog_elements=[],
+            rag_context=None):
```
dmjoy, line 202: Same story here I think (making `rag_context` non-optional)

align_system/algorithms/rag_retreiver_adm_components.py
```diff
@@ -0,0 +1,75 @@
...
+
+        text_splitter = RecursiveCharacterTextSplitter(
+            chunk_size=chunk_size,
+            chunk_overlap=chunk_overlap,
+            add_start_index=add_start_index,
+        )
+        all_splits = text_splitter.split_documents(docs)
+
+        embeddings = HuggingFaceEmbeddings(model_name=embedding_model_name)
+        vector_store = FAISS.from_documents(all_splits, embeddings)
```
dmjoy, line 44: I'm guessing this `vector_store` is just in memory here right?  Or is it backed by a file / cache on disk?  Not requesting any changes here just curious (I know the documents we're using now are small, but what happens if it's a massive collection)

align_system/algorithms/rag_retreiver_adm_components.py
```diff
@@ -0,0 +1,75 @@
...
+from langchain_huggingface import HuggingFaceEmbeddings
+from langchain_core.documents import Document
+
+from align_system.algorithms.abstracts import ADMComponent
+
+DocumentFileType = Union[str, bytes, PathLike]
+DocumentFileListType = Iterable[DocumentFileType]
+
+
+class LangChainRAGIndexerADMComponent(ADMComponent):
```
dmjoy, line 22: I think having RAG implemented as an `ADMComponent` is probably OK for now, but I was thinking of it more like how we have the `StructuredInferenceEngine` pieces (as in the same instance could potentially be re-used by different ADM components).  This _might_ come into play for our multi-attribute ADMs where we need to do some kind of ICL retrieval for relevance computation, and then again for provided in-context examples.

align_system/configs/adm/tagging_fewshot_aligned_rag.yaml
```diff
@@ -0,0 +1,52 @@
...
+  START: ${adm.start}
+  SALT: ${adm.salt}
+  BCD_SIEVE: ${adm.bcd}
+
+step_definitions:
+  rag_indexer:
+    docs_files:
+      - /data/users/yonatan.gefen/align-system/align_system/documents/start.md
+      - /data/users/yonatan.gefen/align-system/align_system/documents/start_triage_flowchart.md
+      - /data/users/yonatan.gefen/align-system/align_system/documents/Salt.md
```
dmjoy, line 34: Better to check with Emily and Aaron here, but I don't know that we want all protocol documents available all the time (it may depend on the target, e.g. for a `start` target, we would use the two start documents, but not the `Salt.md`).

Also we have a convention for storing / using files in the repo for this kind of thing it seems (or at least use a `/data/shared` path rather than your own directory): https://github.com/ITM-Kitware/align-system/blob/main/align_system/configs/adm_component/icl/tagging.yaml#L17-L19

align_system/configs/adm/tagging_rag_baseline.yaml
```diff
@@ -0,0 +1,27 @@
...
+  - /adm_component/misc@step_definitions.ensure_chosen_action: ensure_chosen_action
+  - /adm_component/misc@step_definitions.populate_choice_info: populate_choice_info
+  - _self_
+
+step_definitions:
+  rag_indexer:
+    docs_files:
+      - /data/users/yonatan.gefen/align-system/align_system/documents/start.md
+      - /data/users/yonatan.gefen/align-system/align_system/documents/start_triage_flowchart.md
+      - /data/users/yonatan.gefen/align-system/align_system/documents/Salt.md
```
dmjoy, line 17: Same comment here about filepaths

align_system/prompt_engineering/tagging_prompts.py
```diff
@@ -190,3 +190,82 @@ def __call__(self, target_kdma, target_value):
             return bcd_tagging_system_prompt()
         else:
             raise ValueError(f"Unknown target tagging protocol: {target_kdma}")
+
+
+class SimpleTaggingSystemPrompt:
```
dmjoy, line 195: Slight preference (seems more maintainable) for the `rag_context` based prompts to be separate classes (they can call `super` for `__call__` and then just append the RAG context stuff to keep it a bit neater?

scripts/run-tagging-comparison.sh
dmjoy, line 1: Would prefer these `scripts` to go in the scratch repo

pyproject.toml
```diff
@@ -41,6 +41,12 @@ llama-index-retriever = [
     "langchain>=0.2.5",
     "llama-index>=0.13.0",
 ]
+rag = [
```
dmjoy, line 45: Pros and cons of just including these dependencies in the main list?  (How big / unwieldy is `faiss-cpu` I suspect we already include the others?)

dmjoy (changes requested):
Generally looks pretty good to me, but a handful of requested changes.  I'm waffling a bit on whether or not to include the tagging domain documents in the repo here.  Longer term (as we accumulate more document sets) it doesn't seem sustainable; but it is more convenient and ensures we don't lose the documents we were using.  Probably fine to leave them in for now (as we already include the ICL datasets in the repo).

PR #272, opened 2026-04-27
Added method for prompt generation from raw data

aly-mankowski:
Two phase approach for prompt generation:
1. Extract the relevant information from the raw dataset
2. Simplify extracted entries into a concise list
phase2_direct_gen.yaml shows how this can be configured for medical kdma.


align_system/prompt_engineering/direct_regression_system_prompt.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def __call__(self, model):
+        if self.system_prompt is not None:
+            return self.system_prompt
+
+        # Extract
+        extract_cacher = ub.Cacher(
+            "generated_system_prompt_template_extract",
+            depends=self._cache_repr(self.dataset, self.extract_prompt),
+            enabled=self.enable_caching,
```
dmjoy, line 140: Oh interesting.  So this `enabled` flag turns the caching on or off?  Didn't know that was a feature and could simplify the caching code I have elsewhere.

align_system/algorithms/direct_regression_adm_component.py
```diff
@@ -141,10 +141,19 @@ def run(self,
                     raise RuntimeError(f"Couldn't find character ({action['character_id']}) referenced by action")
 
                 dialog = []
-                if self.per_attribute_templates[attribute].get('system_prompt') is not None:
-                    dialog.insert(0, DialogElement(
-                        role='system',
-                        content=self.per_attribute_templates[attribute]['system_prompt']))
+                system_prompt_template = self.per_attribute_templates[attribute].get('system_prompt_template')
```
dmjoy: Hmm, I think this change (expecting a `system_prompt_template` instead of `system_prompt`) might break some existing configs.  Would much prefer to support both as we do with the baseline component here: https://github.com/ITM-Kitware/align-system/blob/main/align_system/algorithms/outlines_baseline_adm_component.py#L74-L86

align_system/algorithms/direct_regression_adm_component.py
```diff
@@ -141,10 +141,19 @@ def run(self,
...
+                if system_prompt_template is not None:
+                    if callable(system_prompt_template):
+                        system_prompt = call_with_coerced_args(
+                            system_prompt_template,
+                            {'model':self.structured_inference_engine.model}
```
dmjoy: Wondering why we aren't initializing this component with `structured_inference_engine` and instead opting to pass in the model and create a new generator (looks like we're just passing in the `structured_inference_engine.model` anyway).  Are there parameters that need to be different or are we otherwise invoking this differently that makes this necessary?  The main detractor to doing it this way is that we can't freely swap out the inference engine backend.

align_system/prompt_engineering/direct_regression_system_prompt.py
dmjoy, line 1: This file probably makes more sense to go into the `prompt_engineering` directory (instead of `algorithms`)

dmjoy (changes requested):

align_system/algorithms/direct_regression_adm_component.py
aly-mankowski: renamed `system_prompt_template` to `system_prompt` to retain compatibility

dmjoy:
Merging as-is, but when / if we need to revisit we should opt to use the `structured_inference_engine` rather than pass the model instance.

dmjoy (approved):

PR #274, opened 2026-05-15
Conceptagent

Vman11:


align_system/interfaces/ai2thor_env.py
dmjoy, line 1: I feel like maybe this file or the contents of it belongs more in the ai2thor interface component

align_system/algorithms/mcts_adm/itm_adapter.py
dmjoy, line 1: Not certain yet how this piece fits into the whole ADM / pipeline but one though here is that we may not need to conform to the existing ITM scenario state format (as that's program evaluation specific).

align_system/algorithms/mcts_adm/llm_ollama.py
dmjoy, line 1: I think we should just be using the `ollama_inference_engine` here (or ideally any "structured inference engine" so that we can swap them around).  Proposer and critic implementations should initialize with a reference to a structured inference engine to call.

align_system/algorithms/mcts_adm/llm_ollama.py
```diff
@@ -0,0 +1,341 @@
...
+
+        payload = {
+            "task": task,
+            "scenario": obs.text,
+            "available_actions": _tool_compact(tools),
+            "recent_choices": [a.tool_name for a in action_history[-5:]],
+            "k": k,
+        }
+
+        prompt = (
```
dmjoy: We generally store prompts under `align_system/prompt_engineering` or have them be variables rather than hardcoding them.

align_system/algorithms/mcts_adm/llm_ollama.py
```diff
@@ -0,0 +1,341 @@
...
+        except Exception:
+            return 0.0, ""
+
+
+# ---------------------------------------------------------------------------
+# AI2Thor-specific proposer / critic
+# ---------------------------------------------------------------------------
+
+class OllamaAI2ThorProposer(ProposerLLM):
+    """Proposer with prompts tailored for embodied navigation in AI2Thor."""
```
dmjoy: If the only difference here between this and `OllamaITMProposer` then these should just be a single class with the prompt parameterized.

align_system/algorithms/mcts_adm/llm_ollama.py
```diff
@@ -0,0 +1,341 @@
...
+        obs: Observation,
+        tools: List[ToolSpec],
+        action_history: List[Action],
+        k: int,
+        diversity_hint: Optional[str] = None,
+    ) -> List[PlanCandidate]:
+        schema_hint = (
+            '{"candidates":[{"actions":[{"tool_name":"MoveAhead","args":{"moveMagnitude":0.25}}],'
+            '"rationale":"..."}]}'
+        )
```
dmjoy: With our other structured inference engine components we can actually pass in a JSON schema to force the output to be valid (under most conditions)

align_system/algorithms/mcts_adm/mcts_adm.py
```diff
@@ -0,0 +1,83 @@
...
+    log = _logging.getLogger(__name__)
+
+from .itm_adapter import (
+    actions_to_toolspecs,
+    alignment_target_to_task,
+    find_itm_action,
+    state_to_observation,
+)
+from .llm_ollama import OllamaConfig, OllamaITMCritic, OllamaITMProposer
+from .planner import MCTSPlanner
```
dmjoy: I think it's better to use full (module level import "paths") here instead of relative, e.g. `align_system.algorithsm.mcts_adm.itm_adapter`

align_system/algorithms/mcts_adm/mcts_adm.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+class MCTSActionBasedADM(ActionBasedADM):
```
dmjoy: I'm not necessarily opposed to defining a new ADM here but if we can fit this ADM into the `pipeline_adm` (i.e. as a sequence of steps) that would be my preference as there would be a greater chance for component re-use

align_system/algorithms/mcts_adm/mcts_ai2thor_adm.py
```diff
@@ -0,0 +1,101 @@
...
+
+        self.planner = MCTSPlanner(
+            proposer=OllamaAI2ThorProposer(prop_cfg),
+            critic=OllamaAI2ThorCritic(crit_cfg),
+            expansions=expansions,
+            proposals_per_expand=proposals_per_expand,
+            rollout_horizon=rollout_horizon,
+            uct_c=uct_c,
+            seed=seed,
+        )
```
dmjoy: Is the only difference here between `MCTSActionBasedADM` and this one the different proposer and critics being used?  Again seems like there's good overlap between these two where the things that are different could be parameterized.

align_system/algorithms/mcts_adm/planner.py
```diff
@@ -0,0 +1,221 @@
...
+        tools: List[ToolSpec],
+        action_history: List[Action],
+    ) -> List[Action]:
+        root = Node(parent=None, plan_actions= None)
+
+        for _ in range(self.expansions):
+            # 1) selection
+            node = root
+            while node.children and not node.unexpanded:
+                node = self._uct_select(node)
```
dmjoy: I suppose this is maybe why it makes sense not to use the `pipeline_adm` class since the planner is doing this tree based search (so it's not single inference request for propose, followed by single inference request for critique followed by action selection).

align_system/algorithms/mcts_adm/planner.py
```diff
@@ -0,0 +1,221 @@
...
+            if node.visits == 0:
+                return "\033[90m"   # grey
+            if hi == lo:
+                return "\033[92m"   # all equal → just green
+            if node.value >= mid + (hi - mid) * 0.5:
+                return "\033[92m"   # green  — top third
+            elif node.value >= mid:
+                return "\033[93m"   # yellow — middle
+            else:
+                return "\033[91m"   # red    — bottom third
```
dmjoy: We have access to the `rich` library for logging which would probably be better to use here rather than these color codes.

align_system/algorithms/mcts_adm/run_ai2thor.py
dmjoy, line 1: Instead of a dedicated "run" script this should be going through `run_align_system` with hydra configurations setting up everything.

align_system/algorithms/mcts_adm/runner.py
dmjoy, line 1: I'm not sure what this file does?

align_system/algorithms/alignment_adm_component.py
dmjoy, line 1: Not sure why this file is changing, unless there's something weird going on with a merge or rebase (pulling in changes from `main` or another branch).  In general I have a strong preference for rebasing as it keeps the history relatively clean.

align_system/algorithms/argmax_alignment_adm_component.py
dmjoy, line 1: This seems like an appropriate way to implement alignment for your use case.

align_system/algorithms/mcts_candidate_adm_component.py
```diff
@@ -0,0 +1,151 @@
...
+from align_system.algorithms.abstracts import ADMComponent
+from align_system.algorithms.mcts_adm.llm_ollama import OllamaAI2ThorProposer, OllamaConfig
+from align_system.algorithms.mcts_adm.types import Action as MCTSAction, Observation, ToolSpec
+from align_system.interfaces.ai2thor_interface import AI2ThorAction
+from align_system.utils import logging
+
+log = logging.getLogger(__name__)
+
+
+class MCTSCandidateGeneratorADMComponent(ADMComponent):
```
dmjoy: Where does this actually get used?  And similar comments about using a `structured_inference_engine` rather than initializing an ollama connection here.

align_system/algorithms/ollama_inference_engine.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def run_inference(self, prompts, schema: str) -> list[dict]:
+        """
+        Run inference for each prompt string and return parsed JSON dicts.
+
+        `schema` is a JSON Schema string appended to each prompt as an
+        instruction. Malformed responses trigger up to
+        `json_repair_attempts` self-repair calls.
+        """
+        schema_instruction = (
```
dmjoy, line 95: Does Ollama support structured outputs directly or have to provide guidance on the correct schema through prompting?

align_system/algorithms/pipeline_adm.py
dmjoy, line 1: I'm not convinced about these changes though it's an interesting way to do it (there's been some discussion with others on the team about how to manage state / historical information from prior steps).  Do you actually need these things for your implementation?

align_system/configs/adm/pipeline_comp_reg_ai2thor.yaml
dmjoy, line 1: Do these configs actually work? (The `pipeline_comp_reg_ai2thor...` ones)

align_system/configs/attribute/naacl24/attribute_descriptions.yaml
dmjoy, line 1: Confused as to why this file is added here now, or where it came from (thought it already existing somewhere).

align_system/configs/interface/ai2thor.yaml
```diff
@@ -0,0 +1,7 @@
+_target_: align_system.interfaces.ai2thor_interface.AI2ThorInterface
+
+scene: FloorPlan1
+prompts: [0]        # 0=apple, 1=tomato, 2=red fruit
```
dmjoy, line 4: Not sure why it isn't better to have the actual prompts here rather than an index into a variable that lives somewhere else?

align_system/interfaces/ai2thor_interface.py
```diff
@@ -0,0 +1,137 @@
...
+from align_system.algorithms.mcts_adm.ai2thor_env import AI2ThorEnv
+from align_system.algorithms.mcts_adm.types import Action as MCTSAction
+
+
+TASKS = {
+    0: "Pick up an Apple.",
+    1: "Pick up a Tomato.",
+    2: "Pick up the Red Fruit.",
+    3: "URGENT: The stove is on and a loose knife is present — both pose immediate harm risk to a nearby patient. You must act now: turn off the stove knob or pick up the knife to prevent injury.",
+}
```
dmjoy, line 16: I feel like I saw these defined in another file somewhere; do we have 2 kinds of "ai2thor" interface components?

align_system/interfaces/ai2thor_interface.py
```diff
@@ -0,0 +1,137 @@
...
+            unstructured=f"{self.task}\n\n{result.obs.text}",
+            scenario_complete=result.done,
+        )
+        return self._state
+
+    def intend_action(self, action: AI2ThorAction) -> AI2ThorState:
+        return self.take_action(action)
+
+
+class AI2ThorInterface(Interface):
```
dmjoy, line 101: Does anything in this file talk to the AI2Thor simulator as it's running?

pyproject.toml
dmjoy, line 1: Looks like some git conflict stuff got committed here.

dmjoy (changes requested):

align_system/interfaces/ai2thor_env.py
dmjoy: EDIT: Probably better to move this into the interfaces directory (perhaps under a `ai2thor` subdir)

align_system/algorithms/mcts_candidate_adm_component.py
dmjoy: EDIT: This is instantiated via the Hydra configs as a step

align_system/algorithms/mcts_adm/mcts_ai2thor_adm.py
dmjoy: EDIT: This file (and some others) not likely to survive the complete integration; probably safe to ignore

align_system/algorithms/mcts_adm/run_ai2thor.py
dmjoy: EDIT: This file won't be used in the final integration

align_system/algorithms/argmax_alignment_adm_component.py
dmjoy: EDIT: WIP

align_system/algorithms/ollama_inference_engine.py
dmjoy: Ask @alright-code if she knows how to do structured inference with Ollama, I seem to recall maybe yes (when working on the template generation stuff)

align_system/configs/attribute/naacl24/attribute_descriptions.yaml
dmjoy: EDIT: Will be deleted

align_system/algorithms/pipeline_adm.py
dmjoy: @ygefen here's where Vincenzo has implemented some notion of state tracking for the concept agent stuff.  As mentioned above I'm not 100% convinced that this is "the" way to do it, but the footprint is fairly small and there's already some precedent.  What I might change here is having the `update_history` call take in the full `working_output` rather than just the chosen action.

align_system/algorithms/choice_generator_adm_component.py
```diff
@@ -0,0 +1,223 @@
...
+            "from the observation's visible lines (the value after 'id='). "
+            "Never use object type names like 'Apple' as objectId. Full objectIds contain '|' characters.\n"
+            "- Avoid repeating the same last action unless clearly helpful.\n"
+        )
+
+
+        dialog = []
+        dialog.insert(0,DialogElement(content=prompt_system, role="system"))
+        dialog.append(DialogElement(content=prompt, role="user"))
+        dialog.append(DialogElement(content=predict_proposer_prompt, role="user"))
```
dmjoy, line 143: Minor thing, but this seems more clean and readable?
```
dialog = [DialogElement(content=prompt_system, role="system"),
          DialogElement(content=prompt, role="user"),
          DialogElement(content=predict_proposer_prompt, role="user")]
```

align_system/algorithms/pipeline_adm.py
dmjoy: @Vman11 FYSA Yoni is working on a variation of this that I think is a bit more generic.  Once that lands I would like you to update this to use his approach (not expecting a heavy lift there)

align_system/data_models/types.py
dmjoy, line 1: Seems like these types are ai2thor specific, so I'd prefer this file be called something like `ai2thor_types.py` to help namespace them.

align_system/interfaces/ai2thor_interface.py
```diff
@@ -0,0 +1,144 @@
...
+
+from dataclasses import dataclass, field
+from typing import Any, Dict, List, Optional
+
+from align_system.interfaces.abstracts import Interface, ActionBasedScenarioInterface
+from align_system.interfaces.ai2thor_env import AI2ThorEnv
+from align_system.data_models.types import Action as MCTSAction
+
+
+TASKS = {
```
dmjoy, line 11: Again would be good to have this configurable (rather than hardcoded)

dmjoy (changes requested):

PR #277, opened 2026-05-21
Pipeline Statefulness for Open World Challenges

ygefen:


align_system/algorithms/abstracts.py
ygefen, line 1: Addes a `reset_history` default NOOP for backwards compatibility and so that we do not have to check `hasattr` in the code.
In fact I would push to do this same pattern for `output_conflict_resolver` to avoid the hasattr check for that too.

align_system/algorithms/pipeline_adm.py
```diff
@@ -8,8 +8,9 @@
 
 
 class PipelineADM(ActionBasedADM):
-    def __init__(self, steps: list[ADMComponent]):
+    def __init__(self, steps: list[ADMComponent], history_window=None):
         self.steps = steps
+        self.history = deque(maxlen=history_window)
```
ygefen, line 14: `maxlen` controls history window length

align_system/algorithms/pipeline_adm.py
```diff
@@ -21,15 +22,17 @@ def choose_action(self,
...
         per_step_timing_stats = []
 
         for i, step in enumerate(self.steps):
             step_returns = step.run_returns()
 
             start_time = timer()
-            # Run the step
-            run_output = call_with_coerced_args(step.run, working_output)
+            # Run the step, temporarily adding historical working outputs to working_output
+            working_output_with_history = {**{"history": list(self.history)}, **working_output}
```
ygefen, line 36: This is a way to make sure that we don't get recursive nesting of historical working_output in the self.history. We could also delete `working_history[history]` from the dict before appending it to `self.history`, but this seemed cleaner.

align_system/algorithms/pipeline_adm.py
```diff
@@ -73,5 +76,10 @@ def choose_action(self,
...
         working_output.setdefault('choice_info', {})['per_step_timing_stats'] =\
             per_step_timing_stats
-
+        self.history.append(working_output)
         return working_output['chosen_action'], working_output
+
+    def reset_history(self):
+        self.history.clear()
+        for step in self.steps:
+            step.reset_history()
```
ygefen: pipeline steps may retain their own state. e.g. I am thinking about how the prompt creation step may keep a dialog history, extracting the last dialog element from `history[-1]`. Of course that step could also be implemented statelesly if `PipelineADM.history.maxlen = None`, but this gives us flexibility.

align_system/drivers/itm_phase1.py
```diff
@@ -96,9 +96,8 @@ def _compute_time_stats(times_s):
             log.info(f'[bold]*Scenario ID*[/bold]: {scenario.id()}')
 
             # Reset any decision or chat history for a new scenario
-            if hasattr(adm, 'reset_history'):
```
ygefen, line 99: No longer need this LBYL, since we guaranty the existence of this attribute

ygefen (commented):
Explanantions for @dmjoy

align_system/drivers/itm_phase1.py
dmjoy: Call it superstition, but I'd rather leave the check in there.  There's a chance some older ADMs don't implement the `ActionBasedADM` interface (or that some newer ones don't).

align_system/algorithms/abstracts.py
dmjoy: Yeah I agree with you that doing the same thing with `output_conflict_resolver` would tidy some things up; however the way it's set up currently it's configurable at the hydra config level, but this change would only make it configurable at the code level (I think)?  FWIW we don't use this pattern very often but here's one config where it's used: https://github.com/ITM-Kitware/align-system/blob/main/align_system/configs/adm/phase2_pipeline_fewshot_comparative_regression_swap_average.yaml#L53

dmjoy (approved):
I like this implementation (for the same reasons you state).  Only change I'd like to see is to keep the LBYL for `reset_history`.

align_system/algorithms/abstracts.py
ygefen: Cool, so in the ABC it would look like:
```
def output_conflict_resolver(*args, **kwargs):
    return *args, **kwargs
```
i.e. a Callable type noop, and the hydra in your example would override it in the subclass with the injected callable.

align_system/drivers/itm_phase1.py
ygefen: Fine with me, your call.

align_system/algorithms/abstracts.py
dmjoy: Hmm, can you override members like that in the hydra config?  I was under the impression that you can only set / pass initialization args with hydra instantiation.  If the config I linked works without modification with this ABC update you're suggesting then I'm in favor of this change.

PR #280, opened 2026-05-28
added ai2thor enviroment with a method that proposes steps and uses C…

Vman11:
…R to choose steps


align_system/algorithms/ollama_inference_engine.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+class OllamaInferenceEngine(StructuredInferenceEngine):
+    """
+    StructuredInferenceEngine backed by a local Ollama model.
+
+    Unlike the Outlines engine, output is not grammar-constrained — the
+    schema is appended to the prompt as an instruction and the response
+    is parsed as JSON with a repair fallback.
+    """
```
dmjoy: It does look like Ollama supports structured output with a JSON schema: https://ollama.com/blog/structured-outputs.  Seems like we should use that if possible

align_system/algorithms/outlines_inference_engine.py
```diff
@@ -127,40 +132,96 @@ def run_in_batches(
             outputs.extend(output)
         return outputs
 
-    def run_inference(self, prompts, schema):
+    def _parse_json(self, text: str) -> dict:
```
dmjoy: I'm not really excited about the idea of including this here.  Have you been running into JSON validation etc. errors with the outlines inference engine here or?  Just curious why this is even needed.

align_system/algorithms/pipeline_adm.py
dmjoy, line 1: I think it's fine to have the history tracking as you have it in here for now, but I'm more inclined to merge Yoni's approach on this: https://github.com/ITM-Kitware/align-system/pull/277/changes#diff-ea512e45fac46d4935ce85a4837bdbcd27b5891a09c1ac5f6aad038076f4d497

As it maintains the full working_output history.

align_system/algorithms/proposer_adm_component.py
```diff
@@ -0,0 +1,216 @@
...
+            f"Generate {self.num_candidates} diverse candidate plans.\n"
+            f"- Each plan is 1 to {self.rollout_horizon} actions.\n"
+            f"- Use ONLY the tool names provided.\n"
+            f"- Args MUST satisfy each tool schema.\n"
+            f"- IMPORTANT objectId rule: For tools requiring objectId (TeleportNearObject, PickupObject, "
+            f"OpenObject, CloseObject, ToggleObjectOn/Off), you MUST copy the exact full objectId string "
+            "from the observation's visible lines (the value after 'id='). "
+            "Never use object type names like 'Apple' as objectId. Full objectIds contain '|' characters.\n"
+            "- Avoid repeating the same last action unless clearly helpful.\n"
+        )
```
dmjoy: My preference would for these to be done similar to how we do prompts in other ADMs (outlines templates or callables, and parameterize them in the init call so that we can swap them around at Hydra configuration time)

align_system/configs/experiment/ai2thor/planner_comp_reg.yaml
dmjoy, line 1: These experiment configs should be probably be in a subdirectory inside of `experiment` as that's typically how we've been doing it.

align_system/data_models/ai2thor.py
dmjoy, line 1: I feel like these should be in a file specific to AI2Thor (if they are indeed specific data types for that domain) just to help with namespacing.  I.e. if I do `from align_system.data_models.ai2thor import ToolSpec` that seems more informative

align_system/interfaces/ai2thor_env.py
```diff
@@ -0,0 +1,786 @@
...
+        if frame is None:
+            print("[AI2ThorEnv] save_frames=True but event has no frame data")
+            return
+        Path(self.frame_dir).mkdir(parents=True, exist_ok=True)
+        fname = f"step{self._step_count:04d}_{action_name}.png"
+        fpath = os.path.join(self.frame_dir, fname)
+        Image.fromarray(frame.astype(np.uint8)).save(fpath)
+        print(f"[AI2ThorEnv] saved frame: {fpath}")
+
+    def reset(self, task: str) -> Observation:
```
dmjoy: All of the setup info etc. in here seems like it should be living in a data or config file somewhere rather than code?

align_system/interfaces/ai2thor_interface.py
```diff
@@ -0,0 +1,153 @@
...
+
+from dataclasses import dataclass, field
+from typing import Any, Dict, List, Optional
+
+from align_system.interfaces.abstracts import Interface, ActionBasedScenarioInterface
+from align_system.interfaces.ai2thor_env import AI2ThorEnv
+from align_system.data_models.types import Action as PlannerAction
+
+
+TASKS = {
```
dmjoy: Similar story here, this should probably be in a data or configuration file somewhere and we would probably want to be able to modify it for a given experiment.

dmjoy (changes requested):
Overall this is in the right direction, but there are a few things I would like to see changed (see comments) or at least have a good handle on why the current modifications are needed for those cases.

PR #281, opened 2026-06-10
able to config bert scorer for gpu/cpu, as well as cache it

Vman11:


align_system/configs/adm/outlines_regression_aligned/incontext.yaml
```diff
@@ -6,6 +6,8 @@ inference_kwargs:
   incontext:
     number: 5
     method: bert_similarity
+    bert_scorer_device: auto
+    cache_bert_scorer: true
```
dmjoy: I would prefer an implementation where retroactively editing configs isn't needed

align_system/utils/incontext_utils.py
```diff
@@ -25,7 +25,38 @@
...
+    Args:
+        device: Device to run the scorer on ('cpu', 'cuda', etc.).
+                'auto' (or None) selects 'cuda' when available, else 'cpu'
+        cache_scorer: If True, reuse a single scorer instance across calls
+                      to avoid reloading the model
+
+    Returns:
+        BERTScorer instance on the requested device
+    """
+    global _bert_scorer, _bert_scorer_device
```
dmjoy: Thinking we can arrive at a solution that doesn't require globals.  Will include details on my suggested approach in the main review text

align_system/utils/incontext_utils.py
```diff
@@ -25,7 +25,38 @@
...
+        return BERTScorer(lang="en", device=device)
+
+    if _bert_scorer is None or _bert_scorer_device != device:
+        _bert_scorer = BERTScorer(lang="en", device=device)
+        _bert_scorer_device = device
+
+    return _bert_scorer
+
+
+def bert_similarity_selection(candidates, texts_to_compare, reference_text, n_examples, score_adjustments=None, least_similar_examples=False, bert_scorer_device="auto", cache_bert_scorer=True):
```
dmjoy: Similarly I'm not a fan of how we have to now pass around `bert_scorer_device` and `cache_bert_scorer` parameters in several places

dmjoy (changes requested):
I think I'd much prefer the bert scorer / similarly piece to be a proper class / object if possible.  Or is there any reason why we can't initialize the `BERTScorer` instance when the ICL component is initialized (as `self.bert_scorer` or something) with sensible defaults (so they don't need to be specified in most cases) i.e. `device: "auto"`, and then just pass that instance to the `bert_similarity_selection` calls?  Though I'll admit I'm not intimately familiar with the ICL code.  Tagging @eveenhuis as she may have a more informed opinion here.

align_system/utils/incontext_utils.py
dmjoy, line 175: I.e. have a new keyword argument here for `scorer=None`; then in the `__init__` body initialize a BERTScorer (or wrapper object) and set to something like `self.scorer` then pass that to the function for determining matches.

dmjoy (approved):
LGTM!  Did a simple local integration test run to verify.

PR #282, opened 2026-06-11
Phase 2 Eval 2 Training and Observation Runs Configs

dmjoy:
@eveenhuis FYSA


dmjoy:
Just adding configs here.  Merging.

PR #286, opened 2026-07-23
Open World Patient Ranking Pipeline

eveenhuis:
Rebased version of [!283](https://github.com/ITM-Kitware/align-system/pull/283), includes changes from [!273](https://github.com/ITM-Kitware/align-system/pull/273)

See the description from those below:

- Fixes unit tests broken by merit focus weight modification from ADEPT 2026-01-21
- We weren't finding that medical urgency/KDMA(s) varied across actions for the same character.So, we are now regressing based on character descriptions only (or non-character actions, but that doesn't really apply due to the available actions). [OWFormatChoicesADMComponent, then CR/DR]
- We use alignment to pick the highest ranked character (or non-character action) [Existing alignment function].
- That character may correspond to multiple actions, so we then have a secondary non-aligned prompt to choose which sub action to do. [OWChoiceToActionADMComponent]
- Once we've picked a single action, we may have details to fill in like tag color or character ID, so we then do that with followups (again, not aligned) [OWActionParameterCompletionADMComponent]
- Then finally, we do a tag adjustment based on a heuristic to "align" tag color. [OWTaggingAdjustmentADMComponent]. I didn't want to name this component alignment because it doesn't directly use the alignment target. Instead it considers the KDMA ranking vs medical only ranking as well as how other patients have been tagged so far

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 * Added `MultinomialRandomEffectsModelAlignmentADMComponent` for binomial/trinomial alignment based on ADEPT's latest guidance
+* Added `TournamentRandomEffectsModelAlignmentADMComponent`, a tournament-style version of the random effects alignment function
+  (note this is still not multi-kdma)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,14 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 * Added `MultinomialRandomEffectsModelAlignmentADMComponent` for binomial/trinomial alignment based on ADEPT's latest guidance
 * Added `TournamentRandomEffectsModelAlignmentADMComponent`, a tournament-style version of the random effects alignment function
   (note this is still not multi-kdma)
+* Added several new components for "Open World" experimentation: `OWFormatChoicesADMComponent` extracts relevant characters such that
+  the ADM chooses a character instead of an action, `OWChoiceToActionADMComponent` maps a character selection back to an actual action,
+  `OWActionParameterCompletionADMComponent` ensures all required parameters are filled in,`OWTaggingAdjustmentADMComponent` applies
+  heuristics to indirectly affect tag choice by the alignment target based on unaligned vs aligned patient ranking, and
+  `OWRandomParameterCompletionADMComponent` was added so that the random ADM can function in this environment for smoketests.
+* Added a new driver for "Open World" experimentation: `ITMOpenWorldDriver`. Note that while the name implies the ADM can operate in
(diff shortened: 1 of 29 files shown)


align_system/algorithms/open_world_components.py
```diff
@@ -0,0 +1,285 @@
...
+        structured_inference_engine,
+        scenario_description_template,
+        system_prompt=None,
+    ):
+        self.structured_inference_engine = structured_inference_engine
+        self.scenario_description_template = scenario_description_template
+        self.system_prompt = system_prompt
+
+        self.followup_character_prompt = FollowupClarifyCharacterPrompt()
+        self.followup_tag_prompt = FollowupClarifyTagPrompt()
```
dmjoy, line 105: These should probably be parameterized (i.e. added to the `__init__` args)

align_system/configs/experiment/phase2_feb_openworld/phase2_pipeline_direct_regression_random_effects.yaml
```diff
@@ -0,0 +1,45 @@
+# @package _global_
+defaults:
+  - /adm_component/misc@adm.step_definitions.action_parameter_completion: ow_action_parameter_completion
+  - /adm_component/misc@adm.step_definitions.tagging_adjustment: ow_tagging_adjustment
+  - override /adm: phase2_pipeline_direct_regression
+  - override /interface: ta3
+  - override /adm_component/misc@adm.step_definitions.format_choices: ow_format_choices
+  - override /adm_component/alignment@adm.step_definitions.scalar_alignment: tournament_random_effects_tuple
+  - override /adm_component/misc@adm.step_definitions.ensure_chosen_action: ow_choice_to_action
+  - override /driver: itm_phase2_ow
```
dmjoy, line 10: Wondering if it makes more sense to have new `adm` configs (I guess we would just need one for comp reg and one for direct reg) to base these experiment files off of.

CHANGELOG.md
dmjoy, line 1: I suspect there's a few more things we'd want to add to the changelog (new components etc.)

dmjoy (approved):
Mostly minor comments and overall LGTM.  Perhaps most pressing would be to add a bit more to the Changelog, but the other things can be deferred to another PR (or included in a subsequent PR for OW).

align_system/algorithms/open_world_components.py
eveenhuis: When I was writing this component, it felt very specific to this task and so I didn't, but you're right that parameterizing doesn't hurt.

align_system/configs/experiment/phase2_feb_openworld/phase2_pipeline_direct_regression_random_effects.yaml
eveenhuis: Yeah, potentially. The adm directory already feels a little messy so in my mind I was keeping all the open world stuff segregated as much as possible.

CHANGELOG.md
eveenhuis: Added

